PR #10, opened 2018-02-22
Add a number of Critical Section and Critical Region tests

Fisherman166:
These tests all pass on real hardware. WINE unit tests were used as inspiration. KeEnterCriticalRegion fails on Cxbx-Reloaded currently.


Luca1991:
Thank you very much! Your contribution is really appreciated.

PR #11, opened 2018-02-24
When on real hardware, create a logfile

Fisherman166:
On real hardware, only one screen of text is shown at a time. If more than one screens worth of text is printed, then the previous screen is erased, and the new lines are printed on the blank screen. This is not desirable when running a large number of tests. Also, trying to read the text on the console is hard, so being able to look at a logfile back on the computer is nice.

This adds a hook to the print function that also prints the information to a logfile when on the real hardware only. Cxbx-Reloaded does not need since print can be redirected to a kernel.log file already. A logfile named "kernel_test.log" is created in the same directory as the .xbe which can be FTP'd back to your computer for examination.

### DIFF
--- a/main.c
+++ b/main.c
@@ -80,6 +80,8 @@ void main(void){
 
     print("Kernel Test Suite");
 
+    open_output_file("kernel_tests.log");
+
     if(load_conf_file() == 0){
         print("Config File Loaded");
         print("is_emu: %d", is_emu);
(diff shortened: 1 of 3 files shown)


Luca1991:
@Fisherman166 Thanks for your PR. Anyway I'm trying to contact you on Discord. Please login.

PR #16, opened 2018-02-27
Ansi string tests

Fisherman166:
Add basic tests for InitAnsiString and FreeAnsiString


PatrickvL:
Where is assert_ansi_string declared?

PatrickvL:
Ah found it in rtl_assertions, never mind

PR #17, opened 2018-03-02
Some Nt tests

Luca1991:
Some Nt tests:

- NtCreateFile (this might need a second pass)
- NtReadFile
- NtClose

and a deletion to avoid rebooting the xbox at every run (now it just return to the dashboard).

More stuff soon... :tm:

### DIFF
--- a/main.c
+++ b/main.c
@@ -126,5 +126,4 @@ void main(void){
 
     XSleep(10000);
     pb_kill();
-    XReboot();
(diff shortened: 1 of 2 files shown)


PatrickvL:
Thank you! Hopefully you'll be able to expand upon these tests later on (they are rudimentary right now). It's good to see more tests coming in!

PR #21, opened 2018-04-06
More Rtl tests

Luca1991:
Sorry for the delay. So:

 Currently, all AV Tests are disabled due to a mistake that cause cxbx to crash. I'll try to fix the address soon.

Added these Rtl Tests:
* RtlCharToInteger
* RtlCopyString
* RtlEqualString (currently disabled because not working on my real xbox)
* RtlEqualUnicodeString
* RtlFreeUnicodeString
* RtlInitUnicodeString

My next step is to fix the AV tests, the RtlEqualString and then continue with Rtl tests implementation.

PS: Just for reference, Cxbx currently fails on RtlCompareMemoryUlong and RtlEqualUnicodeString

### DIFF
--- a/av_tests.c
+++ b/av_tests.c
@@ -4,34 +4,35 @@
 
 
 void test_AvGetSavedDataAddress(){
-    if(!is_emu){
+/*    if(!is_emu){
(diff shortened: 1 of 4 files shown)


PatrickvL:
RtlCompareMemoryUlong  is fixed now

PR #23, opened 2018-04-07
Add test for RtlCompareMemory

Fisherman166:
Added a simple test for RtlCompareMemory. Also added back in the Av tests that are not causing the crash on Cxbx.

### DIFF
--- a/av_tests.c
+++ b/av_tests.c
@@ -4,20 +4,19 @@
 
 
 void test_AvGetSavedDataAddress(){
-/*    if(!is_emu){
+    if(!is_emu){
          print("0x0001 - AvGetSavedDataAddress: Skipped (due to real hadrware) / value: %x",AvGetSavedDataAddress());
          return;
     }
     AvSetSavedDataAddress((void*) 0x12345678);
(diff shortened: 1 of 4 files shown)


PatrickvL:
Thanks!

Could you also indicate which tests still fail on Cxbx-Reloaded? And possibly attach a dump of the output from a real machine?

Luca1991:
@Fisherman166 @PatrickvL the Av tests, in the current form, causes real hardware to crash. It is better to disable them for now, until I (or Fisherman) fix the address (that should be 0xFD000000 as pointed out by Patrick).

PatrickvL:
... Which should be declared as NV2A_MMIO_BASE

Fisherman166:
test_AvSetDisplayMode is the one that fails on the Cxbx. All of the Av tests are always skipped on real hardware. Luca wrote all of the AV tests, so he would be better suited to to work on fixing them for now.

Luca1991:
Ok good, I'll give it a look ;) Anyway these tests shouldn't be skipped on real hw after the fix.

PR #27, opened 2018-04-09
avoid system crash in Av Tests

Luca1991:
Not really the fix I was hoping for... but at least it doesn't crash on real system anymore and we got rid of these if.


PatrickvL:
Thanks!

PR #29, opened 2018-04-15
Add test for RtlAppendUnicodeStringToString

Fisherman166:
Added a simple test that is similar to RtlAppendStringToString for RtlAppendUnicodeStringToString.


PatrickvL:
Thank you!

PR #33, opened 2018-04-24
Implemented tests for RtlMoveMemory and RtlZeroMemory

Luca1991:
Another two Rtl tests:

- RtlMoveMemory

- RtlZeroMemory

XGetTickCount is used as a rand() replacement, since it is missing from nxdk.

Thanks,
Luca


PatrickvL:
Thanks!

PR #34, opened 2018-04-24
Added tests for RtlLowerChar and RtlUpperChar

Luca1991:
Implemented another two tests:

- RtlLowerChar

- RtlUpperChar

CxBx-Reloaded can't run RtlUpperChar, but don't worry I already know why, and I'm going to fix it, but not now because I have to go. Maybe tonight.

Regards,
Luca


PatrickvL:
Thanks, you're picking up speed, nice!

Still curious to why RtlUpperChar can't run in Cxbx-Reloaded?

PR #39, opened 2018-05-06
Refactor RtlUpperChar again to have 100% coverage

Fisherman166:
I wanted to make sure that we had 100% coverage on this function because the input space is small and one failure has been observed in Cxbx-R already. All inputs were ran on real hardware to obtain results and are used as the expected results in the test.


PatrickvL:
Quite lucky, if you think about it, that we already tested on one out of the total of four failing cases

PR #43, opened 2018-09-01
Add test for RtlUshortByteSwap

Fisherman166:
Tested on Cxbx-R and real hardware with no failures.


PatrickvL:
Thanks!

Strange how the travis-ci build fails...

PR #44, opened 2018-09-01
Add test for RtlUlongByteSwap

Fisherman166:
No failures were observed on real hardware or on Cxbx-R.


PatrickvL:
Keep them coming!

PR #46, opened 2018-10-17
Fix for RtlUpperString tests

### LINKED ISSUE #36: RtlUpperString test hangs on real hardware
When running the kernel test suite on my Xbox, the RtlUpperString tests hangs. The last thing I see printed in the logfile is "0x013D - RtlUpperString: Tests Starting". A hard reset of the console is required to exit. Cxbx-R does not hang when running this test for some reason. @Luca1991 please take a look as you implemented this test.

Please test all changes on real hardware and on Cxbx-R. Pull request #37 comments out the test to make forward progress.

thrimbor:
This fixes the test cases for RtlUpperString. There were three problems with it:
1. res_str was initialized with a string literal, making res_str.Buffer point to a `const char *` which is not allowed for the result buffer.
2. res_str was initialized with the same string literal as src_str, making res_str.Buffer point to the same address as src_str.Buffer.
3. strcmp() was used for the comparison, which won't yield correct results due to RtlUpperString ignoring the null terminator.

I did not run the tests with CxBx-Reloaded, but they worked on real hardware.

Fixes #36.

### DIFF
--- a/rtl_tests.c
+++ b/rtl_tests.c
@@ -1158,77 +1158,76 @@ void test_RtlUpperChar(){
     print_test_footer(func_num, func_name, tests_passed);
 }
 
-// FIXME - This test hangs on real hardware but passes on Cxbx-R
 void test_RtlUpperString(){
-//    const char* func_num = "0x013D";
-//    const char* func_name = "RtlUpperString";
-//    BOOL tests_passed = 1;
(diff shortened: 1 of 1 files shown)


PatrickvL:
Thanks! Much appreciated

PR #47, opened 2018-11-10
Fix RtlEqualString so that it passes on Cxbx-Reloaded and real hardware

Fisherman166:
Real hardware does not clear the eax register before returning the result of RtlEqualString in the al register. To get the real return value on real hardware, I just look at the first bit of the result.

### DIFF
--- a/rtl_tests.c
+++ b/rtl_tests.c
@@ -604,10 +604,7 @@ void test_RtlEqualString(){
 
     BOOLEAN result;
     for(uint8_t i = 0; i < sizeof(str1_inputs) / sizeof(ANSI_STRING*); i++) {
-        // Real hardware does not clear the upper bits of eax in RtlEqualString before setting
-        // the result in the al register. To get the actual result of the function call, only
-        // look at the first bit of the result.
-        result = RtlEqualString(str1_inputs[i], str2_inputs[i], case_insensitive[i]) & 0x1;
+        result = RtlEqualString(str1_inputs[i], str2_inputs[i], case_insensitive[i]);
         if(result == expected_result[i]) {
             print("  Test PASSED for str1 = %s, str2 = %s, case insensitive = %x.",
                   str1_inputs[i]->Buffer, str2_inputs[i]->Buffer, case_insensitive[i]);
--- a/rtl_tests.c
+++ b/rtl_tests.c
@@ -586,30 +586,41 @@ void test_RtlEnterCriticalSectionAndRegion(){
     print_test_footer(func_num, func_name, tests_passed);
 }
 
-void test_RtlEqualString(){ // FIXME this test failed on real hardware!!!
-    /*const char* func_num = "0x0117";
+void test_RtlEqualString(){
+    const char* func_num = "0x0117";
     const char* func_name = "RtlEqualString";
     BOOL tests_passed = 1;
     print_test_header(func_num, func_name);
 
-    ANSI_STRING str1;
-    ANSI_STRING str2;
-    ANSI_STRING str3;
-
+    ANSI_STRING str1, str2, str3;
     RtlInitAnsiString(&str1, "xbox is cool");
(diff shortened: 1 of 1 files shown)


x1nixmzeng:
Interesting! Does this imply that RtlEqualString is returning the wrong number of bits?

Edit: Also worth checking if RtlEqualUnicodeString has the same behaviour

PatrickvL:
Good find! Almost unbelievable how silly this is - could you add a test that proves the EAX upper bits aren't cleared even when the input is equal?

Fisherman166:
@PatrickvL: Adding a test like that would be difficult because we would hit the original issue that caused us to disable this test. Cxbx-R will return 0 or 1, but real hardware will return something like 0x1500 and 0x1501 for the first two test causes. Causing the expected result to be different on each platform.

@x1nixmzeng: The function seems to return eax as I was getting weird values in `BOOLEAN result` like 0x1501. I found this issue by looking at the disassembly of my kernel dump. Eax is used as an upper bound for the character-by-character comparison loop counter. At the end of the function, either `xor al, al` or `mov al, 1` is used to set the return value of the function **without** clearing the value of eax. Checking all xrefs of where this function is called, the instruction immediately following the RtlEqualString call is `test al, al`. Real hardware is expecting to only ever look at the `al` register.

A quick look at RtlEqualUnicodeString has a similar signature. `Eax` is used throughout the function and not cleared before `xor al, al` or `mov al, 1` is used to set the return value.

PatrickvL:
The Cxbx-Reloaded kernel could be updated to reflect this odd behaviour; What value remains exactly in the upper 24 bits of eax?

rtl_tests.c
```diff
@@ -586,30 +586,41 @@ void test_RtlEnterCriticalSectionAndRegion(){
...
+    ANSI_STRING* str2_inputs[]      = {&str1, &str1, &str2, &str2, &str3, &str3};
+    BOOLEAN      case_insensitive[] = {0    , 1    , 0    , 1    , 0    , 1    };
+    BOOLEAN      expected_result[]  = {1    , 1    , 0    , 1    , 0    , 0    };
+
+    BOOLEAN result;
+    for(uint8_t i = 0; i < sizeof(str1_inputs) / sizeof(ANSI_STRING*); i++) {
+        // Real hardware does not clear the upper bits of eax in RtlEqualString before setting
+        // the result in the al register. To get the actual result of the function call, only
+        // look at the first bit of the result.
+        result = RtlEqualString(str1_inputs[i], str2_inputs[i], case_insensitive[i]) & 0x1;
```
JayFoxRox: <https://docs.microsoft.com/en-us/windows/desktop/winprog/windows-data-types> lists it as:
> A Boolean variable (should be TRUE or FALSE). This type is declared in WinNT.h as follows: `typedef BYTE BOOLEAN;`

Asserting for `TRUE` and `FALSE` would be right - looking at the lowest bit is *not* right; you must look at the entire `BYTE`.

JayFoxRox:
I don't understand this issue?

`RtlEqualString` is returning a `BOOL` which is an `unsigned char` in nxdk xboxkrnl.h.
So this can only return values in range 0 - 255 [= same range as `al`].
If this returns anything larger, then it's a bug in nxdk.

*(Independent of this, nxdk also does `typedef unsigend char BOOL` in the same step; which is wrong because `BOOL` should be `int` sized; I also couldn't find any users of `BOOL` in xboxkrnl.h)*

Fisherman166:
Jay was right - it was a compiler bug. I hadn't pulled from nxdk in a while and BOOLEAN was still being typedef'd as an int incorrectly. With the newest nxdk, BOOLEAN is correctly typedef'd as a BYTE, which means this test now passes on Cxbx-R and real hardware without `& 0x1`.

I hadn't even thought of compiler bugs so that is something that I will look out for in the future.

PatrickvL:
Good to know the underlying problem has now been found & fixed, thanks everyone

PR #48, opened 2018-11-13
Use new wchar xlibc function in nxdk for the RtlAppendUnicodeStringToString test

Fisherman166:
After this pull request to nxdk is merged, this pull request can be merged: https://github.com/XboxDev/nxdk/pull/79 

I updated the first unicode string test function to 1. Use the new wchar xlibc functions in nxdk to cleanup p the code and 2. Actually compare the result buffer to an expected unicode string. **This greatly improves testing ability** as before I was just checking if the pointer of the buffer matched itself.

### DIFF
--- a/rtl_tests.c
+++ b/rtl_tests.c
@@ -3,6 +3,7 @@
 #include <string.h>
 #include <stdlib.h>
 #include <ctype.h>
+#include <wchar.h>
 
 #include "output.h"
 #include "common_assertions.h"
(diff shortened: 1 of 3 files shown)


PatrickvL:
May it be assumed this now awaits the merging of https://github.com/XboxDev/nxdk/pull/82 ?

Fisherman166:
Yes that would be the safer bet as that will hopefully be the long-term solution.

PatrickvL:
Status : still awaiting merge of remote PR

rtl_tests.c
```diff
@@ -98,35 +99,36 @@ void test_RtlAppendUnicodeStringToString(){
...
 
-    // wcslen does not work because nxdk fails to find wchar.h
-    const WCHAR src_text[] = L"Xbox";
-    const uint8_t num_chars_in_src = 4;
-    WCHAR buffer[num_chars_in_src * 2];
-    const uint8_t num_buf_bytes = sizeof(buffer);
     UNICODE_STRING src_str, dest_str;
+    RtlInitUnicodeString(&src_str, L"Xbox");
+    const size_t num_chars_in_src = wcslen(src_str.Buffer);
+    WCHAR buffer[(num_chars_in_src + 1) * 2]; // Add 1 to hold the \0 end character
```
x1nixmzeng, line 105: The WCHAR type is already 2 bytes long, there should be no need to multiply by 2

Fisherman166: That code creates a buffer that is 2x the number of wchars (with the NULL terminator \0 taken into account with the +1) as the src. `num_chars_in_src = 4` in this case, so the code creates a buffer that is 10 WCHAR characters large. Big enough to hold src_str.Buffer twice as src_str.Buffer is 5 WCHAR characters large.

JayFoxRox: There should probably be a comment which explains that this will have 2 concatenated copies of the `src_str`. This is indeed very confusing, especially with the zero-termination comment being less relevant [as it's more standard] than the multiplication [which is unique to the use-case].

PatrickvL:
Still awaiting the merge of the reference pr

Fisherman166:
This can finally be merged as my PR was accepted: https://github.com/XboxDev/nxdk/pull/82

PatrickvL:
Status : The remote PR mentioned above is now merged, however, Travis CI now fails building this - once that is addressed, this PR can (finally!) be reviewed & merged.

Fisherman166:
It just needed a rebuild with the new code from nxdk. All passing now.

PR #50, opened 2018-12-07
Implemented tests for most of the crypto functions

ergo720:
This adds tests for the functions XcPKEncPublic, XcPKDecPrivate, XcPKGetKeyLen, XcVerifyPKCS1Signature, XcModExp, XcDESKeyParity, XcKeyTable, XcBlockCrypt, XcBlockCryptCBC and XcCryptService. All tests pass on real hardware and on cxbxr only XcDESKeyParity succeeds because all the other functions are unimplemented (with the exception of XcCryptService, which still succeeds but by mere chance).


PatrickvL:
This must've taken a few hours (days?) to write, well done!

PR #51, opened 2018-12-08
RtlCharToInteger Update

Fisherman166:
The RtlCharToInteger test was missing tests for the '0b', '0o', '0x' base specifiers in the input string. It was also missing tests for invalid numbers for a specified base in the input string, as RtlCharToInteger will convert the valid numbers up to the first invalid number and return the result. Major refactoring of the code has been done to allow future extensions to this function (if needed) to be easier.


PatrickvL:
LGTM!

PR #52, opened 2018-12-16
Rtl extended integer multiply test

Fisherman166:
Add a test for RtlExtendedIntegerMultiply. There isn't much functionality to the function so there isn't much needed to cover it.

I also noticed there is a lot of duplicate code all over for printing the results of tests. To clean that up I added in pass and fail strings that are selected to remove the duplicate printing code.


PatrickvL:
Looks good to me!

PR #55, opened 2019-02-03
Add test for RtlCaptureContext

Fisherman166:
Checks for Ebp, Eip and Esp are missing in the resulting context and will need to be added at a later date.

### DIFF
--- a/rtl_tests.c
+++ b/rtl_tests.c
@@ -211,8 +211,88 @@ void test_RtlAssert(){
     /* FIXME: This is a stub! implement this function! */
 }
 
+// Notes about this function: The official kernel allocates 0x440 bytes for the CONTEXT structure.
+// That does not match with any other implementation of CONTEXT that I have been able to find.
+// RtlCaptureContext does not touch the ContextFlags or Floating point save fields of the CONTEXT
+// structure.
+//
+// FIXME: RtlCaptureContext also sets Ebp, Eip, and Esp in the context record. No checks are enabled for them
+// currently.
 void test_RtlCaptureContext(){
-    /* FIXME: This is a stub! implement this function! */
+    const char* func_num = "0x0109";
+    const char* func_name = "RtlCaptureContext";
+    BOOL tests_passed = 1;
+    print_test_header(func_num, func_name);
+
+    CONTEXT result_context;
+    // cs and ss words so make sure that the upper word of the Seg* dword is 0 or the tests will fail
+    // due to unitialized values.
+    result_context.SegCs = 0;
+    result_context.SegSs = 0;
+
+    DWORD eax_val = 0x336699CC;
+    DWORD ebx_val = 0x11223344;
+    DWORD ecx_val = 0x55667788;
+    DWORD edx_val = 0x99AABBCC;
+    DWORD esi_val = 0xDDEEFF11;
+    DWORD edi_val = 0xDEADBEEF;
+    DWORD flags_val = 0;
+    // Again, cs and ss are words so make sure the upper word is initialized to 0
+    DWORD cs_val = 0;
+    DWORD ss_val = 0;
+    DWORD RtlCaptureContext_address = 0;
+
+    asm volatile(
+        // Make sure we can clobber all registers
+        "pusha\n"
+        // Get address of RtlCaptureContext and store it for later
+        "movl __imp__RtlCaptureContext@4, %%eax\n"
+        "movl %%eax, %0\n"
+        // Push &result_context to pass into RtlCaptureContext
+        "leal %1, %%eax\n"
+        "pushl %%eax\n"
+        // Store current values of cs and ss
+        "leal %2, %%eax\n"
+        "movw %%cs, (%%eax)\n"
+        "leal %3, %%eax\n"
+        "movw %%ss, (%%eax)\n"
+        // Store current value of the flags
+        "leal %4, %%eax\n"
+        "pushf\n"
+        "popl (%%eax)\n"
+        // Setup pre-set register values
+        "movl %5, %%eax\n"
+        "movl %6, %%ebx\n"
+        "movl %7, %%ecx\n"
+        "movl %8, %%edx\n"
+        "movl %9, %%esi\n"
+        "movl %10, %%edi\n"
+        // Call RtlCaptureContext and cleanup registers
+        "calll *%11\n"
+        "popa\n"
+        : "=m" (RtlCaptureContext_address)
+        : "m" (result_context), "m" (cs_val), "m" (ss_val), "m" (flags_val),
+          "m" (eax_val), "m" (ebx_val), "m" (ecx_val), "m" (edx_val), "m" (esi_val),
+          "m" (edi_val), "m" (RtlCaptureContext_address)
+    );
+
+    DWORD* expected_vals[]  = {&eax_val, &ebx_val, &ecx_val, &edx_val, &esi_val, &edi_val,
+                               &cs_val, &ss_val, &flags_val};
+    DWORD* result_vals[]    = {&result_context.Eax, &result_context.Ebx, &result_context.Ecx,
+                               &result_context.Edx, &result_context.Esi, &result_context.Edi,
+                               &result_context.SegCs, &result_context.SegSs, &result_context.EFlags};
+    const char* reg_names[] = {"Eax", "Ebx", "Ecx", "Edx", "Esi", "Edi", "SegCs", "SegSs", "EFlags"};
+
+    for(uint8_t i = 0; i < sizeof(expected_vals) / sizeof(DWORD*); i++) {
+        const char* result_text = passed_text;
+        if(*result_vals[i] != *expected_vals[i]) {
+            tests_passed = 0;
+            result_text = failed_text;
+        }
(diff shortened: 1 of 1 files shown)


rtl_tests.c
```diff
@@ -211,8 +211,80 @@ void test_RtlAssert(){
...
+    asm("movl %0, %%ecx\n" : : "r" (ecx_val));
+    asm("movl %0, %%edx\n" : : "r" (edx_val));
+    asm("movl %0, %%esi\n" : : "r" (esi_val));
+    asm("movl %0, %%edi\n" : : "r" (edi_val));
+
+    // Call RtlCaptureContext and cleanup registers
+    asm("movl __imp__RtlCaptureContext@4, %eax\n");
+    asm("movl %%eax, %0\n" : "=m" (eax_val) : );
+    asm("calll *%eax\n");
+    asm("popa\n");
```
JayFoxRox: Shouldn't all of this code be in one large `asm` block, with the `volatile` keyword?

I'm not sure if the compiler might be allowed to add instructions between ASM blocks, but I'd assume so if it comes to clobbering.

rtl_tests.c
```diff
@@ -211,8 +211,80 @@ void test_RtlAssert(){
...
+    DWORD eax_val = 0;
+    DWORD ebx_val = 0x11223344;
+    DWORD ecx_val = 0x55667788;
+    DWORD edx_val = 0x99AABBCC;
+    DWORD esi_val = 0xDDEEFF11;
+    DWORD edi_val = 0xDEADBEEF;
+    DWORD flags_val = 0;
+    // Again, cs and ss are words so make sure the upper word is initialized to 0
+    DWORD cs_val = 0;
+    DWORD ss_val = 0;
```
JayFoxRox, line 242: Why not store these as `WORD` then?

rtl_tests.c
```diff
@@ -211,8 +211,80 @@ void test_RtlAssert(){
...
+    // Store current values of cs and ss
+    asm("leal %0, %%eax\n" : : "m" (cs_val));
+    asm("movw %cs, (%eax)\n");
+    asm("leal %0, %%eax\n" : : "m" (ss_val));
+    asm("movw %ss, (%eax)\n");
+
+    // Store current value of the flags
+    asm("leal %0, %%eax\n" : : "m" (flags_val));
+    asm("pushf\n");
+    asm("popl (%eax)\n");
```
JayFoxRox: It's trivial to store EBP, and ESP here in a similar fashion.

(even if you don't control them)

rtl_tests.c
```diff
@@ -211,8 +211,80 @@ void test_RtlAssert(){
...
+    asm("movl %0, %%ebx\n" : : "r" (ebx_val));
+    asm("movl %0, %%ecx\n" : : "r" (ecx_val));
+    asm("movl %0, %%edx\n" : : "r" (edx_val));
+    asm("movl %0, %%esi\n" : : "r" (esi_val));
+    asm("movl %0, %%edi\n" : : "r" (edi_val));
+
+    // Call RtlCaptureContext and cleanup registers
+    asm("movl __imp__RtlCaptureContext@4, %eax\n");
+    asm("movl %%eax, %0\n" : "=m" (eax_val) : );
+    asm("calll *%eax\n");
```
JayFoxRox: nit: Could have used a memory location to avoid clobbering EAX.

rtl_tests.c
JayFoxRox: You can add an ASM label here to retrieve the expected EIP address.

rtl_tests.c
```diff
@@ -211,8 +211,80 @@ void test_RtlAssert(){
...
+    asm("popa\n");
+
+    DWORD* expected_vals[]  = {&eax_val, &ebx_val, &ecx_val, &edx_val, &esi_val, &edi_val,
+                               &cs_val, &ss_val, &flags_val};
+    DWORD* result_vals[]    = {&result_context.Eax, &result_context.Ebx, &result_context.Ecx,
+                               &result_context.Edx, &result_context.Esi, &result_context.Edi,
+                               &result_context.SegCs, &result_context.SegSs, &result_context.EFlags};
+    const char* reg_names[] = {"Eax", "Ebx", "Ecx", "Edx", "Esi", "Edi", "SegCs", "SegSs", "EFlags"};
+
+    for(uint8_t i = 0; i < sizeof(expected_vals) / sizeof(DWORD*); i++) {
```
JayFoxRox, line 286: Isn't there an `ARRAY_SIZE(x) sizeof(x) / sizeof(x[0])` macro yet? If not, introduce it?

rtl_tests.c
```diff
@@ -211,8 +211,80 @@ void test_RtlAssert(){
...
+// However, without doing some math there isn't a good way to know what Ebp, Esp, and Eip were
+// set to in the caller of RtlCaptureContext. And I'm too lazy to figure out that code right now.
 void test_RtlCaptureContext(){
-    /* FIXME: This is a stub! implement this function! */
+    const char* func_num = "0x0109";
+    const char* func_name = "RtlCaptureContext";
+    BOOL tests_passed = 1;
+    print_test_header(func_num, func_name);
+
+    CONTEXT result_context;
```
JayFoxRox, line 227: Shouldn't `ContextFlags` be set, or be checked?

Is it an input or output to the function you are testing?

It would probably affect the registers you are allowed to test.

rtl_tests.c
```diff
@@ -211,8 +211,80 @@ void test_RtlAssert(){
     /* FIXME: This is a stub! implement this function! */
 }
 
+// FIXME: RtlCaptureContext also sets Ebp, Eip, and Esp in the context record.
+// However, without doing some math there isn't a good way to know what Ebp, Esp, and Eip were
+// set to in the caller of RtlCaptureContext. And I'm too lazy to figure out that code right now.
```
JayFoxRox: `FloatSave` isn't handled at all. But is not mentioned here, or in the PR description?

rtl_tests.c
Fisherman166: Two reasons: The CONTEXT structure allocates a DWORD for cs and ss, and using a DWORD allows me to use the for loop checking logic at the bottom for these variables.

rtl_tests.c
Fisherman166: This is how the official kernel calls RtlCaptureContext in RtlAssert. It ended up being the only way I could figure out how to make the compiler happy.

rtl_tests.c
Fisherman166: No there is not. Seems like something good to add to nxdk maybe?

rtl_tests.c
Fisherman166: In my kernel dump of RtlCaptureContext, ContextFlags are not touched. CONTEXT addresses 0x21C, 0x218, 0x214, 0x210, 0x20C, 0x208, 0x228, 0x234, 0x22C, 0x224, 0x220, and 0x230 are set in the function which line up with the addresses for the flags and general purpose registers.

rtl_tests.c
Fisherman166: In my kernel dump of RtlCaptureContext, the Floating save area is not touched. CONTEXT addresses 0x21C, 0x218, 0x214, 0x210, 0x20C, 0x208, 0x228, 0x234, 0x22C, 0x224, 0x220, and 0x230 are set in the function which line up with the addresses for the flags and general purpose registers.

rtl_tests.c
JayFoxRox: It shouldn't be part of a toolchain (although it might exist internally) - it should be responsibility of the projects.

rtl_tests.c
JayFoxRox: Probably worth adding a comment here, or even in nxdk.

rtl_tests.c
JayFoxRox: Probably worth adding a comment here, or even in nxdk.

rtl_tests.c
JayFoxRox: Fair enough, although that looks like a bug to me, because they'd destroy EAX, which is probably the most important register.

rtl_tests.c
Fisherman166: Good idea. I will rewrite the section to be one block of volatile asm code.

rtl_tests.c
Fisherman166: After a little bit more experimentation, you were right in that I can call it from memory, which allows me to control eax. It doesn't match how RtlAssert calls RtlCaptureContext, but it adds a bit more testing flexibility.

rtl_tests.c
Fisherman166: Good idea. I'll implement that in a future PR when I get some time.

Fisherman166:
This really got stalled out. What do I need to do to get it moving again?

Fisherman166:
Bumping again.

JayFoxRox:
Nobody in @Cxbx-Reloaded is on the participant list. So they likely don't see your bump unless you use the organization highlight (as I've done). But it often doesn't work either, so you should pick random people from the organization.

@Luca1991 @CakeLancelot @Voxel9 @RadWolfie

PatrickvL:
I suspect it got stalled because it looked like a discussion was still ongoing. A summary might have helped. Having read through the comments it seems this task is done, and some work will be done in a follow-up?

Issue #56, opened 2019-04-24
Unable to build


GXTX:
_Recently_ NXDK moved to a new std library which caused this program to not be able to build.

```
[ LD       ] main.exe
lld: error: /home/wsl/Xbox/SDK/nxdk/samples/xbox_kernel_test_suite/main.obj: undefined symbol: _strtok_r
/home/wsl/Xbox/SDK/nxdk/samples/xbox_kernel_test_suite/../../Makefile:113: recipe for target 'main.exe' failed
make: *** [main.exe] Error 1
```

JayFoxRox:
For the record: the new std lib (pdclib) was in review for weeks, so this didn't come out of nowhere. The new lib is much more complete, stable and cleaner.

The easiest solution is to copy the old source files from nxdk to this project. Ideally you'd stick to standard functions that are part of a libc (`strtok_r` is a POSIX extension) or convince upstream pdclib (https://github.com/DevSolar/pdclib) to change their code.

Luca1991:
Okay, I'll give a look at this issue tomorrow. Are there any other undefined symbols than strtok? 


Edit: to Patrick, and every other cool ppl that was following my work in this repo (and on CxBx-R repo) earlier: sorry for being disappeared that way, I'm still extremely busy completing last couple of exams, following some private projects and looking for a job... I'll be reachable on discord if needed.

Issue #45, opened 2018-10-15
Add Travis CI build link to readme


PatrickvL:
https://travis-ci.org/Cxbx-Reloaded/xbox_kernel_test_suite should be added to the readme

PR #58, opened 2019-04-26
Fix issue #56 and #45 (TEST BEFORE MERGE)

Luca1991:
Hi all,
due to the recent addition of PDClib in nxdk, we weren't able to build xkts due to a missing function. A public domain implementation of this missing function (strtok_r) is now part of xkts. This fix issue #56 . Minor: I also added the Travis CI link to the README, fixing issue #45 .

Now, read carefully:
1) I tested xkts on my real xbox (screenshot: https://i.postimg.cc/Y9HpcH9k/photo-2019-04-26-14-45-39.jpg) and it works, but I was not able to test it using CxBx-R (I can't remember how we used to get the output log...).
2) I found a regression in XCreateFile: it the file already exist, it will NOT overwrite it.

Thank you,
Luca

### DIFF
--- a/string_extra.c
+++ b/string_extra.c
@@ -1,0 +1,33 @@
+#include <stdlib.h>
+#include <string.h>
+
+// Public domain strtok_r() by Charlie Gordon
+char* strtok_r(char *str, const char *delim, char **nextp)
+{
+    char *ret;
+
+     if (str == NULL)
+    {
+        str = *nextp;
+    }
+
+     str += strspn(str, delim);
+
+     if (*str == '\0')
+    {
+        return NULL;
+    }
+
+     ret = str;
+
+     str += strcspn(str, delim);
+
+     if (*str)
+    {
+        *str++ = '\0';
+    }
+
+     *nextp = str;
+
+     return ret;
+}
--- a/README.md
+++ b/README.md
@@ -24,6 +24,9 @@ tests=1,25,3,F
 # BINARIES:
 You can download pre-built bins from here: https://github.com/Cxbx-Reloaded/xbox_kernel_test_suite/releases
 
+# TRAVIS CI:
+Current build status can be seen here: https://travis-ci.org/Cxbx-Reloaded/xbox_kernel_test_suite
+
 # USEFUL LINKS:
 * https://github.com/wine-mirror/wine/tree/master/dlls/ntdll/tests
 * https://github.com/wine-mirror/wine/tree/master/dlls/kernel32/tests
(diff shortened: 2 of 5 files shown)


LukeUsher:
@Luca1991 now that nxdk has a real libc (PDClib) you shouldn't need to use the XCreateFile API anymore. You should be free to use the standard fopen, fread, fwrite, etc family of functions.

JayFoxRox:
Reminder @Cxbx-Reloaded and @Luca1991 

> I found a regression in XCreateFile: it the file already exist, it will now overwrite it.

This change isn't part of this PR? I also don't remember an upstream PR? So why is this mentioned / how was it solved?

string_extra.c
```diff
@@ -0,0 +1,33 @@
+#include <stdlib.h>
+#include <string.h>
+
+// Public domain strtok_r() by Charlie Gordon
+char* strtok_r(char *str, const char *delim, char **nextp)
```
JayFoxRox, line 5: I'm sure I've mentiond this somewhere before, but you can also discuss this upstream at https://github.com/DevSolar/pdclib , so that the pdclib internal functions can possibly expose a re-usable interface.

*(As this would delay this PR for longer, I suggest to still accept this change - then go upstream and try to find out if `__private__strtok_r` might be an upstream-option, to reduce the implementation here)*

Luca1991:
> Reminder @Cxbx-Reloaded and @Luca1991
> 
> > I found a regression in XCreateFile: it the file already exist, it will NOT overwrite it.
> 
> This change isn't part of this PR? I also don't remember an upstream PR? So why is this mentioned / how was it solved?

Hi Jay, no this isn't part of this PR. This is just something I noticed when testing this PR in my real xbox. Also there was a typo in my origianal post, I've written "now" istread of "not". Anyway when I'll have some free time, I'll replace XCreateFile with a standard API as suggested by Luke.

string_extra.c
Luca1991: This is surely a good idea, when I'll have some free time I'll open a ticket in the pdclib repo asking about the possibility to expose strtok_r function.

ergo720:
I tested this with cxbxr https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/commit/1314496cb3d4db0798b11da4a4850c2b668adfa0 and I can confirm that it works fine (but only when building with -DNDEBUG, to workaround another unrelated problem), so this can be merged.

Luca1991:
Just want to let @JayFoxRox know that I opened an issue upstream (pdclib) about the private strtok_r. You can find it here: https://github.com/DevSolar/pdclib/issues/4

PR #59, opened 2019-05-18
video: Add XVideoSetMode()

dracc:
XboxDev/nxdk-pdclib#3 will require XVideoSetMode() to be ran per application.

### DIFF
--- a/main.c
+++ b/main.c
@@ -1,5 +1,6 @@
 #include <xboxrt/debug.h>
 #include <pbkit/pbkit.h>
+#include <hal/video.h>
 #include <hal/xbox.h>
 #include <hal/fileio.h>
 #include <stdio.h>
@@ -92,6 +93,8 @@ static void run_tests() {
 }
 
 void main(void){
+
+    XVideoSetMode(640, 480, 32, REFRESH_DEFAULT);
 
     switch(pb_init()){
         case 0: break;


Luca1991:
Ok, but this should be merged AFTER #58

PatrickvL:
Could someone fix this?
```
[ LD       ] main.exe
lld: warning: /home/travis/build/cxbx-reloaded/nxdk/projects/xbox_kernel_test_suite/main.obj: undefined symbol: _strtok_r
error: link failed
make: *** [main.exe] Error 1
The command "make" exited with 2.
Done. Your build exited with 1.
```

dracc:
@PatrickvL Fixed! :+1:

Luca1991:
@ergo720 I'm not sure that was a good idea to merge this!
The upstream commit (https://github.com/XboxDev/nxdk-pdclib/pull/3) is not yet merged.

EDIT: as Jay pointed out,  

> the existing mode will be set twice (in quick succession)

this shouldn't cause problems, but it wasn't tested as far as I know.

ergo720:
I tested it with https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/commit/1314496cb3d4db0798b11da4a4850c2b668adfa0 and with the workaround mentioned in https://github.com/Cxbx-Reloaded/xbox_kernel_test_suite/issues/60: it works fine for me (I should probably have mentioned it though).

Issue #14, opened 2018-02-25
XCreateFile fail to open config.txt


Luca1991:
For some unknown reason XCreateFile (inside load_conf_file function) is unable to open the config.txt.

I remember this working the last time I checked (first days of Jan). I'll give it a look soon.

Fisherman166:
As a temporary fix on the real hardware it is possible to just open "config.txt" and it will look for config.txt in the same directory as the .xbe. Pull Request #15 does this. This solution does not work on the emulator however.  This piece of code is the source of the issue because it returns an empty string:

+    // FIXME - On my computer, file_path = "" when I run this code
+    //strcpy(file_path, getCurrentDirString());
+    //strcat(file_path, "config.txt");

PatrickvL:
Does that mean `strcat` doesn't work?

Luca1991:
I'm not at home right now, but reading the @Fisherman166 I remembered that `getCurrentDirString()` is an internal nxdk function (that I copied from `fileio.c`), and shouldn't be called directly. Anyway I'll fix this today afternoon.

Luca1991:
This should be fixed in PR #19 

sorry for taking so long....

PatrickvL:
Does it work now on both real machine and Cxbx-Reloaded?

Luca1991:
Sorry I've only tested on my real xbox. I'm on linux, so testing in cxbx might be a bit difficult for me right now.

Fisherman166:
My hunch is that it will not work on Cxbx-Reloaded. Once an XBE starts, the filesystem seen by the XBE is the virtual filesystem created by Cxbx, not the host file system. Unless the user has their test suite XBE stored in the internal filesystem, I don't see how the path to look for the config file inside of the test suite would match.

Also, when building with the latest commit, linking fails for me. I also see warnings during compile, which did not exist before.

lld-link: error: /mnt/e/Windows/Programming/nxdk/my_progs/xbox_kernel_test_suite/main.obj: undefined symbol: _strtok_r
lld-link: error: /mnt/e/Windows/Programming/nxdk/my_progs/xbox_kernel_test_suite/main.obj: undefined symbol: _strtok
lld-link: error: /mnt/e/Windows/Programming/nxdk/my_progs/xbox_kernel_test_suite/main.obj: undefined symbol: _strtol
[...]

PR #61, opened 2020-01-08
Update depreciated nxdk functions

Fisherman166:
JayFoxRox mentioned in [this](https://github.com/Cxbx-Reloaded/xbox_kernel_test_suite/issues/14) issue that we needed to stop using depreciated nxdk functions. I have updated all of the depreciated functions to use the equivalent new functions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,22 +1,20 @@
 language: c
 compiler: gcc
-dist: trusty
+dist: bionic
 os: linux
 notifications:
     email: false
 
 before_install:
-    # add LLVM repository
-    - wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key|sudo apt-key add -
-    - echo "deb http://apt.llvm.org/trusty/ llvm-toolchain-trusty-7 main" | sudo tee -a /etc/apt/sources.list
     - sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test
     - sudo apt-get update -qq
     
 install:
     # install required packages
-    - sudo apt-get install build-essential flex bison clang git clang-7 lldb-7 llvm-7 llvm-7-dev -y
+    - sudo apt-get install build-essential flex bison clang git clang-8 lldb-8 llvm-8 llvm-8-dev -y
     # create an alias for lld-link
-    - alias lld-link=lld-link-7.0
+    - alias lld-link=lld-link-8.0
+    - alias clang=clang-8
     
 before_script:
     # clone NXDK repository to /home/travis/build/Cxbx-Reloaded/
--- a/main.c
+++ b/main.c
@@ -2,8 +2,6 @@
 #include <pbkit/pbkit.h>
 #include <hal/video.h>
 #include <hal/xbox.h>
-#include <winapi/fileapi.h>
-#include <winapi/handleapi.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
@@ -33,6 +31,7 @@ int load_conf_file(char *config_file_path) {
     DWORD file_size = GetFileSize(handle, NULL);
     if(file_size == INVALID_FILE_SIZE) {
         print("ERROR: Could not get file size for %s", config_file_path);
+        return -1;
     }
 
     char* buffer = (char *)malloc(file_size);
--- a/main.c
+++ b/main.c
@@ -1,11 +1,13 @@
-#include <xboxrt/debug.h>
+#include <hal/debug.h>
 #include <pbkit/pbkit.h>
 #include <hal/video.h>
 #include <hal/xbox.h>
-#include <hal/fileio.h>
+#include <winapi/fileapi.h>
+#include <winapi/handleapi.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <windows.h>
 #include "output.h"
 #include "global.h"
 #include "func_table.h"
@@ -13,40 +15,40 @@
 #include "string_extra.h"
 
 int load_conf_file(char *config_file_path) {
-    int handle;
-    unsigned int file_size = 0;
-    char *buffer;
-    unsigned int read = 0;
-    int result;
-
     print("Trying to open config file: %s", config_file_path);
-
-    result = XCreateFile(&handle,
+    HANDLE handle = CreateFile(
         config_file_path,
         GENERIC_READ,
         FILE_SHARE_READ,
+        0,
         OPEN_EXISTING,
-        FILE_ATTRIBUTE_NORMAL);
-    if(result != 0) {
+        FILE_ATTRIBUTE_NORMAL,
+        NULL
+    );
+    if(handle == INVALID_HANDLE_VALUE) {
         print("Could not open config file '%s' for read", config_file_path);
         return -1;
     }
 
-    XGetFileSize(handle, &file_size);
+    DWORD file_size = GetFileSize(handle, NULL);
+    if(file_size == INVALID_FILE_SIZE) {
+        print("ERROR: Could not get file size for %s", config_file_path);
+    }
 
-    buffer = malloc(file_size);
+    char* buffer = (char *)malloc(file_size);
     if(buffer == NULL) {
         print("Malloc failed for file_size %u", file_size);
         return -1;
     }
 
-    result = XReadFile(handle, buffer, file_size, &read);
-    if(result == 0 || read != file_size) {
-        print("Read failed for config file. result = %d, read = %u", file_size, read);
+    DWORD bytes_read = 0;
+    BOOL result = ReadFile(handle, buffer, file_size, &bytes_read, NULL);
+    if(result == 0 || bytes_read != file_size) {
+        print("Read failed for config file. result = %d, read = %u", file_size, bytes_read);
         return -1;
     }
 
-    XCloseHandle(handle);
+    CloseHandle(handle);
 
     char *line;
     char *rest = buffer;
@@ -99,7 +101,7 @@ void main(void){
     switch(pb_init()){
         case 0: break;
         default:
-            XSleep(2000);
+            Sleep(2000);
             XReboot();
             return;
     }
@@ -117,6 +119,6 @@ void main(void){
     vector_free(&tests_to_run);
     close_output_file();
 
-    XSleep(10000);
+    Sleep(10000);
     pb_kill();
 }
--- a/output.c
+++ b/output.c
@@ -1,7 +1,6 @@
 #include <pbkit/pbkit.h>
 #include <hal/debug.h>
-#include <winapi/fileapi.h>
-#include <winapi/handleapi.h>
+#include <windows.h>
 #include <stdio.h>
 #include <stdarg.h>
 #include <stdlib.h>
@@ -10,7 +9,7 @@
 #include "global.h"
 #include "output.h"
 
-static HANDLE output_filehandle = NULL;
+static HANDLE output_filehandle = INVALID_HANDLE_VALUE;
 
 void print(char* str, ...){
     va_list args;
@@ -78,7 +77,7 @@ int write_to_output_file(void* data_to_print, DWORD num_bytes_to_print) {
     }
 
     DWORD bytes_written;
-    int ret = WriteFile(
+    BOOL ret = WriteFile(
         output_filehandle,
         data_to_print,
         num_bytes_to_print,
--- a/output.c
+++ b/output.c
@@ -1,6 +1,7 @@
 #include <pbkit/pbkit.h>
-#include <xboxrt/debug.h>
-#include <hal/fileio.h>
+#include <hal/debug.h>
+#include <winapi/fileapi.h>
+#include <winapi/handleapi.h>
 #include <stdio.h>
 #include <stdarg.h>
 #include <stdlib.h>
@@ -9,7 +10,7 @@
 #include "global.h"
 #include "output.h"
 
-static int output_filehandle = 0;
+static HANDLE output_filehandle = NULL;
 
 void print(char* str, ...){
     va_list args;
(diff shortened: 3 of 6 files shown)


rtl_tests.c
```diff
@@ -1320,8 +1316,8 @@ void test_RtlUpperString(){
     char rnd_letter;
     char rnd_letters[101];
 
-    for(int k=0; k<100; k++){ // we use XGetTickCount as a rand() replacement
-        rnd_letter = "abcdefghijklmnopqrstuvwxyz"[(int)XGetTickCount() % 26];
+    for(int k=0; k<100; k++){ // we use GetTickCount as a rand() replacement
+        rnd_letter = "abcdefghijklmnopqrstuvwxyz"[(int)GetTickCount() % 26];
```
JayFoxRox, line 1320: This is independent of this PR, but this is *bad*. The CPU runs at 733MHz, so 1.36425648 **nano**seconds per cycle.
GetTickCount is in **milli**seconds. So when this loop runs the value will probably be constant.

It's also not giving consistent values when running the test multiple times.

You should switch to a constant seed and a random function. pdclib implements `srand` and `rand`; it also implements `rand_s` but that can't be seeded, so it's not recommended here.

output.c
```diff
@@ -1,6 +1,7 @@
 #include <pbkit/pbkit.h>
-#include <xboxrt/debug.h>
-#include <hal/fileio.h>
+#include <hal/debug.h>
+#include <winapi/fileapi.h>
+#include <winapi/handleapi.h>
```
JayFoxRox: I personally just use windows.h (which is also suggested on the MSDN pages).

The winapi include folder is an implementation detail. If nxdk samples do this differently *I* would consider that a minor bug (although other developers might disagree).

I think it's possible that we'll eventually move files so it's `#include <fileapi.h>` (as it would be on Windows).. or this stops working completely (with only `#include <windows.h>` working).

main.c
```diff
@@ -1,52 +1,54 @@
...
+    );
+    if(handle == INVALID_HANDLE_VALUE) {
         print("Could not open config file '%s' for read", config_file_path);
         return -1;
     }
 
-    XGetFileSize(handle, &file_size);
+    DWORD file_size = GetFileSize(handle, NULL);
+    if(file_size == INVALID_FILE_SIZE) {
+        print("ERROR: Could not get file size for %s", config_file_path);
```
JayFoxRox, line 33: If this error is hit, the next `malloc` would likely fail. You should probably `return -1`

main.c
```diff
@@ -1,52 +1,54 @@
-#include <xboxrt/debug.h>
+#include <hal/debug.h>
 #include <pbkit/pbkit.h>
 #include <hal/video.h>
 #include <hal/xbox.h>
-#include <hal/fileio.h>
+#include <winapi/fileapi.h>
+#include <winapi/handleapi.h>
```
JayFoxRox: Should not be necessary as you have windows.h below

JayFoxRox:
~~There is a typo in the commit message~~ It's actually just british english. Don't mind me.

output.c
```diff
@@ -9,7 +10,7 @@
 #include "global.h"
 #include "output.h"
 
-static int output_filehandle = 0;
+static HANDLE output_filehandle = NULL;
```
JayFoxRox: A better value would be `INVALID_HANDLE_VALUE`

output.c
```diff
@@ -49,38 +50,40 @@ void print_test_footer(
...
-int write_to_output_file(void* data_to_print, unsigned int num_bytes_to_print) {
+int write_to_output_file(void* data_to_print, DWORD num_bytes_to_print) {
     if(is_emu) {
         return 0;
     }
 
-    unsigned int bytes_written;
-    int ret = XWriteFile(
+    DWORD bytes_written;
+    int ret = WriteFile(
```
JayFoxRox: This should return a `BOOL`

ergo720:
The Travis build is failing with the error `use of undeclared identifier 'char16_t/char32_t'`, so this cannot be merged until that is fixed. Additionally, the `GetTickCount` issue is still there, but that should probably be corrected in a separate PR.

Fisherman166:
I was getting the char16_t char32_t errors as well before updating my clang version locally. I've never touched Travis or any other CI before so I'm hesitant to make the change, but my guess would be we need to add this to the install: section:

` - alias clang=clang-7`

JayFoxRox:
Yes, sounds like a clang issue. That proposed fix could work - why wasn't it tried?

You are also running Travis CI on Ubuntu 14.04 which only gets security updates by now.
You should probably switch to 18.04 (which has higher default package versions). While nxdk isn't too cutting edge in terms of OS requirements, I assume that we'll pay less attention to 16.04 reports once 20.04 hits. 4 years is a long time in software development, and it's usually free and easy to upgrade.

If you want to overhaul your CI, you could probably also just steal / mirror the nxdk GitHub Actions workflow. I assume it only requires minimal changes, and you could always just adapt the latest upstream changes (thereby avoiding bitrot).

JayFoxRox:
For your information @Fisherman166 :
- It's probably clang-7.0 (not just 7) in the `alias`.
- `     - echo "deb http://apt.llvm.org/trusty/ llvm-toolchain-trusty-7 main" | sudo tee -a /etc/apt/sources.list ` needs to be updated as you have moved away from trusty.

Fisherman166:
Xenial with alias clang=clang-7 passed build so it seems to be working.

JayFoxRox:
You should still avoid listing trusty package sources on non-trusty Ubuntus. See http://apt.llvm.org/

Xenial (16.04) is also almost 4 years old, so don't expect support to last too long.
Also it looks like [Bionic has pre-packaged clang-8 already](https://packages.ubuntu.com/search?suite=bionic&searchon=names&keywords=clang). So the apt would only be necessary for nightly builds probably. Xenial is stuck on pre-packaged clang-6.

.travis.yml
```diff
@@ -1,22 +1,23 @@
...
 os: linux
 notifications:
     email: false
 
 before_install:
     # add LLVM repository
-    - wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key|sudo apt-key add -
-    - echo "deb http://apt.llvm.org/trusty/ llvm-toolchain-trusty-7 main" | sudo tee -a /etc/apt/sources.list
+    #- wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key|sudo apt-key add -
+    #- echo "deb http://apt.llvm.org/trusty/ llvm-toolchain-trusty-7 main" | sudo tee -a /etc/apt/sources.list
```
ergo720: Shouldn't this be deleted instead of just being commented out?

Fisherman166: I could have but I figured I'd leave an example of how to do it in case someone needed to re-add it in the future. Definitely not needed.

ergo720: There are plenty of tutorials on the Internet that teach how to fetch a gpg key and add a repository to your Linux distro, I wouldn't pollute the code with this. So, I still think you should just remove it.

JayFoxRox: git has a history (and a blame or diff view quickly reveals why this apt code was removed). This github PR also highlights the old code and the discussion about it.

Keeping dead code is not helpful in my opinion (it's just confusing).

Fisherman166: Fair enough. I've deleted it.

ergo720:
I tested these changes with https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/commit/08691c24e17f48e7cecc839fb4a7fd6543d332c5 and compared the test results with the build with tag 20190604141115-c3dc177 and confirmed that they are identical, so this doesn't seem to cause any regressions. I will merge this in a day or so if there are no more remarks.

JayFoxRox:
I'm surprised to hear that it still works given https://github.com/XboxDev/nxdk-pdclib/pull/15#issuecomment-575432934. I don't think it should be able to open the config file (or the output file) anymore?
The Xbox kernel used in our winapi implementation shouldn't be able to load a relative path (whereas Windows probably allows it).

Please retest that on hardware; I believe it only works due to Cxbx-R bugs.

I'm personally not too hyped about hardcoding `D:\\` as the application directory (particularly in the libc API). For the winapi I'd consider that more standardish (also because winapi itself already makes some other assumptions about the platform the code will run on, whereas libc is supposed to be portable). So I can't really recommend that as permanent solution (yet), but we also don't have another solution right now either.

Anyhow, I don't think there are many filepaths in Cxbx-R. You should keep an eye on how https://github.com/XboxDev/nxdk-pdclib/pull/15 develops and potential follow-ups in the nxdk winapi lib.

Also a nit about the PR: Commit titles on this repo typically don't end in a period.

Fisherman166:
This is what it looks like running on CXBX-R:

> [0x36EC] INFO : DBG     Trying to open config file: config.txt
> [0x36EC] INFO : DBG     Could not open config file 'config.txt' for read
> [0x36EC] INFO : DBG     Kernel Test Suite: Skipping creating kernel_tests.log because on emulator
> [0x36EC] INFO : DBG     Kernel Test Suite
> [0x36EC] INFO : DBG     Running tests in emulator mode
>No Specific tests specified. Running all tests (Single Pass).

This is what it looks like on real hardware:

> 
> Kernel Test Suite
> Running tests in real hardware mode
> No Specific tests specified. Running all tests (Single Pass).

This works because in globals.c we set is_emu = 1 by default. On CXBX-R the config file fails to load but it works fine because is_emu is already set to 1. On real hardware the config file can be opened, so we create a config file that sets is_emu=0 which allows the test suite to work in both environments. Its not a great solution, but its a workaround for the filesystem differences that exist between real hardware and CXBX-R.

ergo720:
I'm unable to reproduce your issue, for me it still works fine, even with is_emu=0. I retested with https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/commit/08691c24e17f48e7cecc839fb4a7fd6543d332c5 and a config.txt with is_emu=0 and no tests specified, and the following are the relevant kernel calls extracted from the kernel log generated by Cxbx-R.

```
[0x2704] INFO : DBG     Trying to open config file: config.txt
[0x2704] DBG     DbgPrint returns 0
[0x2704] NT      NtCreateFile forwarding to "IoCreateFile"...
[0x2704] IO      IoCreateFile(
 OUT FileHandle         : 0x149BFE00
   DesiredAccess        : 80100080
   ObjectAttributes     : 0x149BFDE4 -> OBJECT_ATTRIBUTES* {
   .RootDirectory        : FFFFFFFD
   .ObjectName           : 0x149BFDF8 -> STRING* {
   .Length               : A
   .MaximumLength        : B
   .Buffer               : (char *)0x000397B0 = "config.txt"}
   .Attributes           : 40}
 OUT IoStatusBlock      : 0x149BFDF0
   AllocationSize       : 0x00000000
   FileAttributes       : 80
   ShareAccess          : 1
   Disposition          : (CREATE_DISPOSITION)0x00000001 = FILE_OPEN
   CreateOptions        : (CREATE_OPTION)0x00000020 = FILE_SYNCHRONOUS_IO_NONALERT
   Options              : 0
);
[0x2704] DEBUG: IO      IoCreateFile = 0x00000648
[0x2704] IO      IoCreateFile returns 0
[0x2704] NT      NtQueryInformationFile(
   FileHandle           : 00000648
 OUT IoStatusBlock      : 0x149BFDD8
 OUT FileInformation    : 0x149BFDA0
   Length               : 38
   FileInformationClass : (FILE_INFORMATION_CLASS)0x00000022 = FileNetworkOpenInformation
);
[0x2704] NT      NtQueryInformationFile returns 0
[0x2704] NT      NtReadFile(
   FileHandle           : 00000648
   Event                : 00000000
   ApcRoutine           : 00000000
   ApcContext           : 00000000
 OUT IoStatusBlock      : 0x149BFE00
 OUT Buffer             : 0x14B505F8
   Length               : 8
   ByteOffset           : 0x00000000
);
[0x2704] NT      NtReadFile returns 0
[0x2704] NT      NtClose(
   Handle               : 00000648
);
[0x2704] NT      NtClose returns 0
[0x2704] NT      NtCreateFile forwarding to "IoCreateFile"...
[0x2704] IO      IoCreateFile(
 OUT FileHandle         : 0x149BFE4C
   DesiredAccess        : 40100080
   ObjectAttributes     : 0x149BFE30 -> OBJECT_ATTRIBUTES* {
   .RootDirectory        : FFFFFFFD
   .ObjectName           : 0x149BFE44 -> STRING* {
   .Length               : 10
   .MaximumLength        : 11
   .Buffer               : (char *)0x0003A0E7 = "kernel_tests.log"}
   .Attributes           : 40}
 OUT IoStatusBlock      : 0x149BFE3C
   AllocationSize       : 0x00000000
   FileAttributes       : 80
   ShareAccess          : 2
   Disposition          : (CREATE_DISPOSITION)0x00000005 = FILE_OVERWRITE_IF
   CreateOptions        : (CREATE_OPTION)0x00000020 = FILE_SYNCHRONOUS_IO_NONALERT
   Options              : 0
);
[0x2704] DEBUG: IO      IoCreateFile = 0x00000648
[0x2704] IO      IoCreateFile returns 0
[0x2704] DBG     DbgPrint(
   Format               : (char *)0x000399A2 = "%s"
   "..."                : ...
);
[0x2704] INFO : DBG     Kernel Test Suite
[0x2704] DBG     DbgPrint returns 0
[0x2704] NT      NtWriteFile(
   FileHandle           : 00000648
   Event                : 00000000
   ApcRoutine           : 00000000
   ApcContext           : 00000000
 OUT IoStatusBlock      : 0x149BFC14
   Buffer               : 149BFC94
   Length               : 12
   ByteOffset           : 0x00000000
);
[0x2704] NT      NtWriteFile returns 0
[0x2704] DBG     DbgPrint(
   Format               : (char *)0x000399A2 = "%s"
   "..."                : ...
);
[0x2704] INFO : DBG     Running tests in real hardware mode
[0x2704] DBG     DbgPrint returns 0
[0x2704] NT      NtWriteFile(
   FileHandle           : 00000648
   Event                : 00000000
   ApcRoutine           : 00000000
   ApcContext           : 00000000
 OUT IoStatusBlock      : 0x149BFBF8
   Buffer               : 149BFC78
   Length               : 24
   ByteOffset           : 0x00000000
);
[0x2704] NT      NtWriteFile returns 0
[0x2704] DBG     DbgPrint(
   Format               : (char *)0x000399A2 = "%s"
   "..."                : ...
);
[0x2704] INFO : DBG     No Specific tests specified. Running all tests (Single Pass).
```

On Cxbx-R, file creation is trapped by `NtCreateFile`, which simply forwards to `IoCreateFile` that does the actual work. As you can see from above, both calls for config.txt and kernel_tests.log are succeeding, and I find the kernel_tests.log file in the directory of the test suite xbe with the test results written to it.

RadWolfie:
@Fisherman166 I can confirm latest Cxbx-Reloaded build is able to read `config.txt` file just fine.

Few things could cause this to happen:
A) Cxbx-Reloaded doesn't have permission to access the file (file was created/saved as admin)
B) didn't create the file in same directory where loaded default.xbe is at
C) using older cxbxr build (buggy build)
D) extended "255" filepath length limitation

Cheers

JayFoxRox:
I'm surprised this got merged. I'm assuming this broke real hardware support (despite what @Fisherman166 claimed).

- I still believe this will not work on a real MS kernel because that doesn't support relative paths (as shown in the comment I linked above, to warn about this). **`CreateFile` with a path of "foobar" will *not* work on Xbox; it needs to be "D:\\\\foobar"**. OpenXDK `XCreateFile` had a special handler for this, nxdk winapi does not (but it *might* be added in the future, see the linked discussion).
- I still believe that it does work in Cxbx-R (because of inaccurate emulation / missing restrictions in Cxbx-R). **`CreateFile` with a path of "foobar" will work on Windows**. I'm assuming that Cxbx-R maps the Xbox `CreateFile` function to the Windows `CreateFile` function without explicitly adding the Xbox specific restriction; so your code probably still works in Cxbx-R.

RadWolfie:
`XCreateFile` is depreciated from https://github.com/XboxDev/nxdk/pull/150 pull request. It would be better to create an issue or make a pull request to correct this issue for xbox hardware.

---
I can confirm on xbox hardware with real MS kernel isn't able to read the `config.txt` file. Since it has been output to screen, no output to a log file, and even set to one test (to see earlier messages) also didn't change the output result.

rtl_tests.c
RadWolfie: Moved into issue report: https://github.com/Cxbx-Reloaded/xbox_kernel_test_suite/issues/62

PR #63, opened 2020-01-25
Fix relative path usage bug

### LINKED ISSUE #14: XCreateFile fail to open config.txt

ergo720:
This is intended to fix the relative path bug issue introduced in https://github.com/Cxbx-Reloaded/xbox_kernel_test_suite/pull/61. I already tested this on both my Xbox and with https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/commit/08691c24e17f48e7cecc839fb4a7fd6543d332c5 and I can confirm that they are both able to open `config.txt` and write `kernel_tests.log` files.

Closes https://github.com/Cxbx-Reloaded/xbox_kernel_test_suite/issues/14

### DIFF
--- a/output.c
+++ b/output.c
@@ -49,15 +49,15 @@ void print_test_footer(
     }
 }
 
-void open_output_file(char* file_name) {
+void open_output_file(char* output_file_path) {
     if(is_emu) {
-        print("Kernel Test Suite: Skipping creating %s because on emulator", file_name);
+        print("Kernel Test Suite: Skipping creating %s because on emulator", output_file_path);
         return;
     }
 
-    debugPrint("Creating file %s", file_name);
+    debugPrint("Creating file %s", output_file_path);
     output_filehandle = CreateFile(
-        file_name,
+        output_file_path,
         GENERIC_WRITE,
         FILE_SHARE_WRITE,
         0,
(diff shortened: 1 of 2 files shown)


output.c
```diff
@@ -49,15 +49,15 @@ void print_test_footer(
     }
 }
 
-void open_output_file(char* file_name) {
+void open_output_file(char* output_file_path) {
```
RadWolfie: Include `output_` to the argument isn't needed when the function name is `open_output_file`. Plus there's only one file path input to the function. However, `file_path` is better than `file_name`.

JayFoxRox:
~~This PR should probably be marked for closing #14? (the previous PR #61 aimed to do it, but actually didn't?)~~ *This was added to the PR description.*

output.c
ergo720: I changed `output_file_path` to `file_path` and also `config_file_path` to `file_path` in `load_conf_file` or else the name of the argument of the two functions would become inconsistent (which is also why I did this change in the first place).

RadWolfie (approved):

PR #69, opened 2021-04-01
Replace Travis with GitHub Actions CI

Margen67:


Luca1991:
Awesome! if this is working, it is a great replacement for Travis.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,57 @@
...
+          submodules: recursive
+      - name: Set xbox_kernel_test_suite dir
+        id: test
+        run: echo "::set-output name=dir::projects/xbox_kernel_test_suite"
+      - name: Checkout xbox_kernel_test_suite
+        uses: actions/checkout@v2
+        with:
+          path: ${{ steps.test.outputs.dir }}
+      - name: Install dependencies
+        run: echo "/usr/lib/llvm-10/bin" >> $GITHUB_PATH
```
RadWolfie: Can you add a comment for the reason of need for this? Along with the side disclaimer of the number need an update over time.
Also, include `https://github.com/XboxDev/nxdk/issues/196` in the comment as a reference for known issue. Which apparently is not resolved.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,57 @@
...
+      - name: Build
+        working-directory: ${{ steps.test.outputs.dir }}
+        run: make -j$(nproc)
+      - name: Set artifact+release parameters
+        id: parameters
+        run: |
+          export file=default.xbe
+          echo "::set-output name=name::$file"
+          echo "::set-output name=path::${{ steps.test.outputs.dir }}/bin/$file"
+          echo "::set-output name=tag::$(date +'%Y%m%d%H%M%S')-$(git log --format=%h -1)"
```
RadWolfie: For whatever the reason... Format's hash is wrong base on your forked releases page.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,57 @@
...
+      - uses: actions/upload-artifact@v2
+        with:
+          name: ${{ steps.parameters.outputs.name }}
+          path: ${{ steps.parameters.outputs.path }}
+          if-no-files-found: error
+      - uses: softprops/action-gh-release@e99df7cd8088d9c5dddab02bf975f62025a6e760
```
RadWolfie: I still don't agree with using someone else's action for the release process. You can use our existing Cxbx-Reload's CI method which use `hub release` to upload artifact(s).

RadWolfie (changes requested):

.github/workflows/ci.yml
```diff
@@ -0,0 +1,57 @@
...
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout nxdk
+        uses: actions/checkout@v2
+        with:
+          repository: XboxDev/nxdk
+          submodules: recursive
+      - name: Set xbox_kernel_test_suite dir
+        id: test
```
RadWolfie: Instead of make id "test", you could use "xkts" instead. 😉

.github/workflows/ci.yml
RadWolfie: image

Could you somehow explain what happened to your releases page?

RadWolfie:
While having Github's Actions put into use internally is great thing to have. I can foreseen side effect if I, approved, merge this of duplicate releases.

There are few of review remarks that need some changes at the moment.

.github/workflows/ci.yml
Margen67: `create-release` was recently marked as unmaintained: https://github.com/actions/create-release/issues/119

Margen67:
> 
> 
> While having Github's Actions put into use internally is great thing to have. I can foreseen side effect if I, approved, merge this of duplicate releases.
> 
> There are few of review remarks that need some changes at the moment.

The only way(s) to fix that are:
  1. Comment out Travis' deploy.
  ~~2. Remove Travis completely.~~ done

Luca1991:
I think we can remove Travis deploy, as GitHub Actions is actually more integrated and better.

.github/workflows/ci.yml
Margen67: See https://github.com/XboxDev/nxdk/issues/196#issuecomment-812238967

.github/workflows/ci.yml
Margen67: I don't know why that happens. `git rev-parse --short HEAD` or `${GITHUB_SHA:0:7}` could probably be used instead.

.github/workflows/ci.yml
RadWolfie: > Archive this repository, putting it read-only mode. **It will continue to work for all existing Action users, but won't receive any updates.**

Like it said, it will continue work for existing Action users. We don't need any new features from it. `hub release` for artifact uploads are more than enough for us.

.github/workflows/ci.yml
RadWolfie: I'm not sure about `${GITHUB_REF:0:7}` since it may not have a "merged" commit hash? It needs some testing though.

.github/workflows/ci.yml
Margen67: The problem with it being unmaintained is that if a vulnerability is found it won't be fixed.
Is there a way to create a release with the same body, etc as the `create-release` action?

RadWolfie:
```yaml
        run: |
          export tag=$(git rev-parse --short HEAD)
          export tag_full=$(git rev-parse HEAD)
          hub release create -p -m "CI-${tag}" -t ${tag_full} ${tag}
```
^ is proof of concept that appear to work. Turns out Create-Release action's empty body content will cause GitHub's backend to automatic fill the body with commit's message.
refs:
- https://github.com/actions/create-release/blob/main/src/create-release.js#L19
- https://github.com/actions/create-release/blob/main/src/create-release.js#L39-L48

Do note if we add another -m will be going into release's body. First -m argument is the name of the release's title.

On side note, with combination of files attachement. We can do this in one execution. Array of files may need to be done in a loop as usual.

**EDIT:** Do note there's `-t "<branch/commitish>"` which will enforce to use the target instead of default branch.

Margen67: I got release creation working with `gh`:
```
      - name: Create release
        if: |
          github.event.action != 'pull_request' &&
          github.ref == 'refs/heads/master'     &&
          github.repository == 'Cxbx-Reloaded/xbox_kernel_test_suite'
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
        run: gh release create ${{ steps.parameters.outputs.tag }} ${{ steps.parameters.outputs.path }} --target $GITHUB_REF -R $GITHUB_REPOSITORY
```

.github/workflows/ci.yml
Margen67: I made a typo. GITHUB_REF is the branch, not SHA.
I ended up using `${GITHUB_SHA:0:7}`.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,60 @@
...
+          path: ${{ steps.parameters.outputs.path }}
+          if-no-files-found: error
+      - name: Create release
+        if: |
+          github.event.action != 'pull_request' &&
+          github.ref == 'refs/heads/master'     &&
+          github.repository == 'Cxbx-Reloaded/xbox_kernel_test_suite'
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: gh release create ${{ steps.parameters.outputs.tag }} ${{ steps.parameters.outputs.path }} --target $GITHUB_REF -R $GITHUB_REPOSITORY
```
RadWolfie: https://github.com/Cxbx-Reloaded/xbox_kernel_test_suite/pull/69#discussion_r638211036
Like you said `GITHUB_REF` is a branch name. By doing this way, you are going to introduce an issue. If multiple pull requests merged at same time will tagged be at same commit as latest branch's. Please replace `GITHUB_REF` to `GITHUB_SHA` where it should make a release on specific commit. (This is intended design for create-release's action.)

As for `-R $GITHUB_REPOSITORY`, for what reason do we need it?

.github/workflows/ci.yml
RadWolfie: review remark continue at: https://github.com/Cxbx-Reloaded/xbox_kernel_test_suite/pull/69#discussion_r654692645

.github/workflows/ci.yml
```diff
@@ -0,0 +1,60 @@
...
+        with:
+          path: ${{ steps.xkts.outputs.dir }}
+      - name: Install dependencies
+        run: |
+          sudo apt-get -y update
+          sudo apt-get -y install lld llvm
+      - name: Build
+        working-directory: ${{ steps.xkts.outputs.dir }}
+        run: make -j$(nproc)
+      - name: Set artifact+release parameters
```
RadWolfie:
```suggestion
      - name: Set artifact + release parameters
```
or 
```suggestion
      - name: Set artifact/release parameters
```
for readability reason

.github/workflows/ci.yml
Margen67: `-R $GITHUB_REPOSITORY` is needed because otherwise it tries to create a release on the main repository which is problematic for testing on forks.

Margen67:
Travis is broken and isn't free anymore so I've removed it.

.github/workflows/ci.yml
RadWolfie: Base on gh's repository info, `GITHUB_REPOSITORY` is not used by default. Instead, it is checking for `GH_REPO` environment. I recommend to use the environment set than pass in argument. The reason I preferred in environment setter is to remain noticable. Until official github's gh tool proceed integrate support for `GITHUB_REPOSITORY` environment.

Margen67: Using `env:`  (https://github.com/cli/cli/issues/3556) doesn't work for some reason:
```
        env:
          GH_REPO: ${{ env.GITHUB_REPOSITORY }}
```
https://github.com/Margen67/Test/runs/2868128291#step:2:6

RadWolfie: I made a test to be sure if result is correct. Turns out all `GITHUB_` prefix is intended to be run from run command and excluded from env context unless provided one within workflow file. Proper solution is to use 
```
        env:
          GH_REPO: ${{ github.repository }}
```

RadWolfie (approved):

.github/workflows/ci.yml
Margen67: Correction: It tries to create a release on XboxDev/nxdk because it's the workspace. Under normal circumstances overriding the repo like this isn't necessary.
The target, however, always needs to be overridden because `gh release create` doesn't use the current branch.

PR #78, opened 2023-05-16
Refactor files listing for better organization 

### LINKED ISSUE #3: Consider splitting tests into one file per API
Currently, tests are grouped together according to kernel prefix (Ob, Ex, Mm, etc).

It's probably better to create a separate test file per API, because otherwise there's a real chance some test files will become too big. Also, having separate test file per API results in a cleaner revision history.

This comes at a cost though:
* The current setup has to be split up
* Any change in testing method later on in the lifetime of this project, will require updating 379 files

This last point makes it even more important to get the testing framework "first-time-right" (see issue #2).

RadWolfie:
Follow up changes to according ticket #3 expectation. If any changes request for how directories/files should flow. Then please use the ticket instead of here, mainly to continue discussion there.

On side remark, there is no template folder for reusable code established yet.

close #3


PatrickvL (approved):
Great work!

RadWolfie:
@ergo720 found a typo in the commit message. It has been corrected from nativiation to navigation.

PR #79, opened 2023-05-17
Add complete set of Ke's Deferred Procedure Call tests

RadWolfie:
I had verified on hardware for passing state. However, from Cxbx-Reloaded is failing for incorrect procedure but did not crash. Even after I add FS offset of 0x58 patch for writer support, it does not pass the test anyway due to host's KPCR is stored per thread.

One other thing I notice about Cxbx-Reloaded's FS patches, it does not offer full set of patches for each FS offset which can be a problematic. Cxbx-Reloaded currently has reader support but does not have writer support which may break Windows until a FS patch is implemented.

I made couple of TODOs that are in need a fix.
1) There's missing kernel defines in nxdk that are in need of implement. For now, I extend the define to avoid conflict if they are upstream into nxdk.
2) FIXED ~~I tried my best to make AT&T syntax for read/write FS offset instructions. But kept having problems with compiler, even though it looks correct to me according to the documentations.~~
   - Thanks to @ergo720's eyes on nxdk's exception does have usage for FS offset instruction found at: https://github.com/XboxDev/nxdk/blob/5a0af6887fde082c24b1f78804afaf1a373eff6e/lib/xboxrt/vcruntime/excpt.cpp#L65

### DIFF
--- a/src/tests/ke/suite/dpc.c
+++ b/src/tests/ke/suite/dpc.c
@@ -1,0 +1,23 @@
+#include <xboxkrnl/xboxkrnl.h>
+
+#include "util/output.h"
+
+void test_KeInitializeDpc()
+{
+    /* FIXME: This is a stub! implement this function! */
+}
+
+void test_KeInsertQueueDpc()
+{
+    /* FIXME: This is a stub! implement this function! */
+}
+
+void test_KeIsExecutingDpc()
+{
+    /* FIXME: This is a stub! implement this function! */
+}
+
+void test_KeRemoveQueueDpc()
+{
+    /* FIXME: This is a stub! implement this function! */
+}
--- a/src/tests/ke/suite/dpc.c
+++ b/src/tests/ke/suite/dpc.c
@@ -30,6 +30,7 @@ static void setDpcRoutineActive(ULONG setter)
 
 static KDPC dpcObject;
 static BOOL dpc_called;
+static ULONG dpc_executing;
 static void __stdcall dpc_callback(
 	IN KDPC* Dpc,
 	IN PVOID DeferredContext,
@@ -40,6 +41,11 @@ static void __stdcall dpc_callback(
     // DPC called variable
     if (SystemArgument1 && (DWORD)SystemArgument1 > 0xFF) {
         *(BOOL*)SystemArgument1 = TRUE;
+    }
+
+    // Request KeIsExecutingDpc return value
+    if (SystemArgument2 && (DWORD)SystemArgument2 > 0xFF) {
+        *(ULONG*)SystemArgument2 = KeIsExecutingDpc();
     }
 }
 
@@ -174,7 +180,56 @@ void test_KeInsertQueueDpc()
 
 void test_KeIsExecutingDpc()
 {
-    /* FIXME: This is a stub! implement this function! */
+    const char* func_num = "0x0079";
+    const char* func_name = "KeIsExecutingDpc";
+    BOOL test_passed = 1;
+    print_test_header(func_num, func_name);
+
+    // Initialize test
+    KeInitializeDpc(&dpcObject, dpc_callback, NULL);
+    KDPC restore_dpc_object = dpcObject;
+
+    // Test direct call without additional steps
+    dpc_executing = KeIsExecutingDpc();
+    GEN_CHECK(dpc_called, FALSE, "dpc_called");
+    GEN_CHECK(dpc_executing, FALSE, "dpc_executing");
+
+    // Reset test
+    dpc_called = FALSE;
+    dpc_executing = FALSE;
+
+    // Test insertion & executed DPC callback within KeInsertQueueDpc function
+    BOOLEAN inserted = KeInsertQueueDpc(&dpcObject, &dpc_called, &dpc_executing);
+    GEN_CHECK(dpc_called, TRUE, "callback.dpc_called");
+    GEN_CHECK((dpc_executing != FALSE), TRUE, "callback.dpc_executing");
+
+    // Reset test
+    dpc_called = FALSE;
+    dpc_executing = FALSE;
+
+    // Test manual raise to verify if it thinks is running in DPC part 1
+    KIRQL oldIrql = KfRaiseIrql(DPC_DISPATCH_LEVEL);
+    dpc_executing = KeIsExecutingDpc();
+    GEN_CHECK(dpc_called, FALSE, "dpc_called");
+    GEN_CHECK(dpc_executing, FALSE, "dpc_executing");
+    KfLowerIrql(oldIrql);
+
+    // Reset test
+    dpc_called = FALSE;
+    dpc_executing = FALSE;
+
+    // Test manual raise to verify if it thinks is running in DPC part 2
+    oldIrql = KeRaiseIrqlToDpcLevel();
+    dpc_executing = KeIsExecutingDpc();
+    GEN_CHECK(dpc_called, FALSE, "dpc_called");
+    GEN_CHECK(dpc_executing, FALSE, "dpc_executing");
+    KfLowerIrql(oldIrql);
+
+    // Reset test
+    dpc_called = FALSE;
+    dpc_executing = FALSE;
+
+    print_test_footer(func_num, func_name, test_passed);
 }
 
 void test_KeRemoveQueueDpc()
--- a/src/tests/ke/suite/dpc.c
+++ b/src/tests/ke/suite/dpc.c
@@ -9,6 +9,7 @@
 
 typedef struct _dpc_test {
     BOOLEAN inserted;
+    BOOLEAN removal;
     BOOL dpc_called;
     KDPC dpcObject;
 } dpc_test;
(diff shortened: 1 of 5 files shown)


src/tests/ke/suite/dpc.c
```diff
@@ -0,0 +1,279 @@
...
+
+static void setDpcRoutineActive(ULONG setter)
+{
+    __asm__ ("movl %0, %%fs:0x58"
+            :
+            : "a"(setter));
+}
+
+static KDPC dpcObject;
+static BOOL dpc_called;
```
JayFoxRox, line 33: This, and all variations of it, must probably be some kind of atomic or marked `volatile`, because it could change anytime, without the compiler knowing about it.

src/tests/ke/suite/dpc.c
```diff
@@ -0,0 +1,279 @@
...
+static ULONG dpc_executing;
+static void __stdcall dpc_callback(
+	IN KDPC* Dpc,
+	IN PVOID DeferredContext,
+	IN PVOID SystemArgument1,
+	IN PVOID SystemArgument2
+)
+{
+    // DPC called variable
+    if (SystemArgument1 && (DWORD)SystemArgument1 > 0xFF) {
```
JayFoxRox, line 43: What is this check for `> 0xFF`?
= Why are there random integers being passed to this?

RadWolfie: They are tested for SystemArgument# from DPC's object structure. I assumed you didn't notice 0x#0 from KeInsertQueueDpc calls? Therefore should not be used to set a pointer's value from the callback ***if*** the callback did get triggered. For example... Cxbx-Reloaded doesn't support single* KPCR across all threads and will produce non-fatal failure rather than fatal crash isn't what we want.

src/tests/ke/suite/dpc.c
RadWolfie: How is it needed when interrupt level will call dpc callback once KfLowerIrql is called? See example of test code what to expect:
https://github.com/Cxbx-Reloaded/xbox_kernel_test_suite/blob/570f45d652a487eaa4b5515a2a780c345f0d8046/src/tests/ke/suite/dpc.c#L140-L149
Basically higher IRQL's threads will start if has interrupt requested which give them higher priority to run them first before return to original thread.
- Test 3c is expected to be false even if sleep call been placed there.
- Same thing for test 3d as well.
- So, it is always test 3e will return true no matter what.

One other thing I want to add if you noticed blocks of `GEN_CHECK` is moved to bottom of the test. That is due to getting early triggered unexpected behavior from higher IRQLs lowering down to DISPATCH_LEVEL then to original thread via doing print to screen and/or write to file.

Finally, xbox kernel test suite is only running from one thread. Each test is required to create separate thread and destruction only if specific test needs it. For these tests here, it is not needed. Mainly because xbox kernel will call dpc callback once IRQL is lowering down to original IRQL stored by us from KfRaiseIRQL function.

PatrickvL: The meaning and function of various "magic values" may be clear to authors, but not to readers. Even authors may forget this and might need to research them when revisiting older code.

One way to avoid that, is to declare symbol names for each value  (perhaps static constant variables or compiler defines), along with an accurate and concise comment to explain their existence with.

RadWolfie: Each part of the tests already have comments before them. Nothing require anything special from compiler nor constant variables. The main reason these global variables are outside of the function that are being tested is due to they should not be on local stack of current thread as switch to different thread, such as dpc thread, is likely will corrupt the data from their own thread. This is something about readers should at least understand when dealing with multi-threading environment.

Also, it's unnecessary to have such as atomic to make it safe anyway due to KfLowerIRQL call will trigger higher IRQL threads for any interrupt has been requested. Which then will eventually return to owner of the thread.

Perhaps read the comments first for each block of test code before making this conclusion? If there's something not understandable context, please do make review remarks on them instead of here.

Anyway, dpc_ prefix variables are meant to read values from dpc_callback for the test. Which is why they are kept together in same spot. Including the reason of having them as global variables.

RadWolfie: I'm marking them as invalid due to insufficient reading the whole tests.

src/tests/ke/suite/dpc.c
RadWolfie: I'm marking them as invalid due to insufficient reading the whole tests.

PatrickvL: My other comment was meant for this line, where there IS an unexplained `0xFF` value being tested against (and another one 3 lines down).

Besides there not being an expansion on why exactly the value `0xFF` is present/needed here, the first part of the check could be removed (and similarly 3 lines down), because if the second check succeeds, the first part is always `true` too and thus unnecessary.

src/tests/ke/suite/dpc.c
```diff
@@ -0,0 +1,279 @@
...
+        *(BOOL*)SystemArgument1 = TRUE;
+    }
+
+    // Request KeIsExecutingDpc return value
+    if (SystemArgument2 && (DWORD)SystemArgument2 > 0xFF) {
```
PatrickvL, line 48: One possible reason why the value `0xFF` is used here, might be because a test fills a struct with `0xCC` bytes. (If this is the reason, isn't that resulting in `0xCCCCCCCC`, invalidating the assumption?)

Perhaps the above is nonsense (hard to tell, without an explanatory comment, which is kind of the whole point of these remarks), an alternative explanation  might be that this `0xFF` is used to avoid accessing the zero page (the first 4 kilobytes of memory), which would otherwise trigger page faults. (If this is the reason, it might be better to use `PAGE_SIZE` instead). But this too might not be the intent...

src/tests/ke/suite/dpc.c
PatrickvL: Btw, it's fine you don't agree with these remarks, but it's inappropriate to conclude reviewers 'not reading the whole code' - perhaps we did and still don't understand why this magic value is here for. Why else would two experienced coders independently conclude a certain review remark is appropriate?

Reviews are meant to show where a submitted pull request / code changes could be made even better. (Because it's great to see these API's tested! Nothing but praise on that!)

src/tests/ke/suite/dpc.c
RadWolfie: https://github.com/Cxbx-Reloaded/xbox_kernel_test_suite/blob/570f45d652a487eaa4b5515a2a780c345f0d8046/src/tests/ke/suite/dpc.c#L96
https://github.com/Cxbx-Reloaded/xbox_kernel_test_suite/blob/570f45d652a487eaa4b5515a2a780c345f0d8046/src/tests/ke/suite/dpc.c#L101-L102

---

https://github.com/Cxbx-Reloaded/xbox_kernel_test_suite/blob/570f45d652a487eaa4b5515a2a780c345f0d8046/src/tests/ke/suite/dpc.c#L111
https://github.com/Cxbx-Reloaded/xbox_kernel_test_suite/blob/570f45d652a487eaa4b5515a2a780c345f0d8046/src/tests/ke/suite/dpc.c#L115-L116

---

https://github.com/Cxbx-Reloaded/xbox_kernel_test_suite/blob/570f45d652a487eaa4b5515a2a780c345f0d8046/src/tests/ke/suite/dpc.c#L128
https://github.com/Cxbx-Reloaded/xbox_kernel_test_suite/blob/570f45d652a487eaa4b5515a2a780c345f0d8046/src/tests/ke/suite/dpc.c#L160-L161
https://github.com/Cxbx-Reloaded/xbox_kernel_test_suite/blob/570f45d652a487eaa4b5515a2a780c345f0d8046/src/tests/ke/suite/dpc.c#L166-L167

---

The above points out that KeInsertQueueDpc function has SystemArgument# which will be set in the structure when inserted. It is half mainly to check if dpc's object structure does get set properly or not. Or the structure provided by nxdk could be wrong, at least it is correct.

It has nothing to do with test_KeInitializeDpc function that sets 0xCCs since it doesn't insert into dpc queue line.
EDIT: It had been answered to the question at <https://github.com/Cxbx-Reloaded/xbox_kernel_test_suite/pull/79#discussion_r1199609899>

src/tests/ke/suite/dpc.c
RadWolfie: See comment below review remark of its intentional design.

src/tests/ke/suite/dpc.c
PatrickvL: Thanks for all the copy-pastes, I've read all that quite a few times before I commented. It's obvious the tests check passed-in arguments are indeed stored where expected. That's good.

Still, the check against 0xFF in `dpc_callback` comes across as arbitrarily, until all tests are investigated. For instance the second argument is once passed in as an actual memory address, to test that `KeInsertQueueDpc` indeed triggers the dpc. Other times it's just used as a recognizable argument value that should not perform that "is the dpc triggered" check.
This all boils down to:
* the suggestion to explain in a comment why the 0xFF is used like this,
* perhaps replacing all magic values with named symbols, 
* Perhaps use `PAGE_SIZE` instead of 0xFF (or create and use a tooling function like `IsValidPointer(PVOID)`)
* Reducing the two comparisons into one

RadWolfie: I can add additional comment except I'm not planning to replace any of magic values with named symbols at all. Since I preferred to do manual increment for each test performed and the tests are really that small. I can see the point for reduce two comparisons into one. Originally it came as `if (SystemArgument2) {`, then noticed they are stored into the structure. Which had me add `SystemArgument2 > 0xFF` comparison check. I don't plan on use PAGE_SIZE since I am only performing the test within one byte size. Which has nothing to do with memory nor relevant to paging size. As for `IsValidPointer(PVOID)`, I preferred to avoid any kernel APIs, nor exceptions, contacts or otherwise will make the test invalid unintentionally. I'm pretty sure we can't have any unhandled exceptions in dpc thread anyway.

Such example of using nxdk's print to screen and/or write to log file can unintentionally break the test results. Why is that? There are couple manual hacks in place to force mimic how kernel APIs would work.

TLDR edition:
I'll add a comment and reduce comparison to one for both SystemArgument# inputs.

RadWolfie: If happy with the review remarks in #80 pull request. Please mark this conversation as resolved and approve the changes there.

PR #80, opened 2023-05-22
review remark fixes for DPC series tests

RadWolfie:
These review remarks fixes are from #79 pull request:
- global variables clarification
- shrink two comparisons to one
- add comment for 0xFF comparison check

With those review remarks aside, I noticed there's indent mistakes which are fixed here. And change the input values to align with test's number which give more room to add further tests if needed.

If there are any further review remarks in #79, please do it there. Any changes will be reflect to here.


PatrickvL (approved):
Looks good to me, thanks

PR #81, opened 2023-11-24
Added tests for MmAllocateSystemMemory and MmFreeSystemMemory

ergo720:
Note that I have only confirmed these tests with [nboxkrnl](https://github.com/ergo720/nboxkrnl). So, before merging this, someone should probably check that they work as expected on a real xbox too.

### DIFF
--- a/src/tests/mm/MmAllocateSystemMemory.c
+++ b/src/tests/mm/MmAllocateSystemMemory.c
@@ -1,18 +1,7 @@
 #include <xboxkrnl/xboxkrnl.h>
 
 #include "util/output.h"
-
-#define PAGE_SIZE 4096
-
-#define PAGE_NOACCESS          0x01
-#define PAGE_READONLY          0x02
-#define PAGE_READWRITE         0x04
-#define PAGE_EXECUTE           0x10
-#define PAGE_EXECUTE_READ      0x20
-#define PAGE_EXECUTE_READWRITE 0x40
-#define PAGE_GUARD             0x100
-#define PAGE_NOCACHE           0x200
-#define PAGE_WRITECOMBINE      0x400
+#include "assertions/defines.h"
 
 #define SYSTEM_MEMORY_BASE    0xD0000000
 #define SYSTEM_MEMORY_SIZE    (512 * 1024 * 1024) // = 0x20000000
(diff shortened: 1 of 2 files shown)


ergo720:
After a lengthy discussion with @RadWolfie,

src/tests/mm/MmAllocateSystemMemory.c
```diff
@@ -1,8 +1,58 @@
...
 
 void test_MmAllocateSystemMemory()
 {
-    /* FIXME: This is a stub! implement this function! */
+    const char* func_num = "0x00A7";
+    const char* func_name = "MmAllocateSystemMemory";
+    BOOL test_passed = 1;
+    print_test_header(func_num, func_name);
+
+   // NOTE: these tests will be marked as failed if not enough memory can be allocated for them
```
RadWolfie:
```suggestion
   // NOTE: these tests could be marked as failed if not enough memory can be allocated for them
```

src/tests/mm/MmFreeSystemMemory.c
```diff
@@ -1,8 +1,34 @@
...
+
+    // Test #1 - free one page
+    PVOID sys_addr = MmAllocateSystemMemory(PAGE_SIZE, PAGE_READONLY);
+    GEN_CHECK(sys_addr != NULL, TRUE, "SystemMemoryAddress (RO - PAGE_SIZE)");
+    if (sys_addr) {
+        num_of_pages = MmFreeSystemMemory(sys_addr, 0);
+        GEN_CHECK(num_of_pages, 1, "num_of_pages (RO - PAGE_SIZE)");
+    }
+
+    // Test #1 - free two pages
```
RadWolfie:
```suggestion
    // Test #2 - free two pages
```

src/tests/mm/MmAllocateSystemMemory.c
RadWolfie: For clarification, if memory allocated failed. The following reasons could be the cause:
- Not enough memory to allocated (previous xbe used up all the memory and didn't freed them up)
- defective chipset (hardware issue)
- stub function which only return null pointer. (custom kernel issue, not an issue with the test itself)
- etc

PatrickvL:
Why was this closed without merging?

ergo720:
> Why was this closed without merging?

Because I and RadWolfie had different views on the PR, and we couldn't agree on it.

PatrickvL:
Sad

PR #83, opened 2023-11-29
Add support for devcontainer locally and remotely

RadWolfie:
I had not thoroughly tested this template for visual studio code. There could be some unnecessary extension, intellisense issues, etc.

However, this template had been my goto devcontainer for other nxdk relative projects. Feel free to find any faults and note them here. And feel free to create codespace, from GitHub, to try them out from the web browser. Please remember to delete your codespace creation after being done with it.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,15 @@
+FROM ghcr.io/xboxdev/nxdk:latest
+
+# NOTE: curl-dev is require to workaround unknown option failure from libcurl (assumed is old version)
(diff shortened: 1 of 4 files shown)


RadWolfie:
I recently made an discovery for intelliSenseMode need `windows-` prefix in order to resolve `L"string"` misunderstanding wide char is not 4 bytes but actually 2 bytes long.

Other than that, the current issue is XBAPI/NTAPI/CDECL/FASTCALL/DECLSPEC_NORETURN are producing false positive errors. Unless they were replaced with `__stdcall` and the like. I'm not sure about DECLSPEC_NORETURN's issue to resolve this.

PR #88, opened 2023-12-26
Added tests for MmAllocateSystemMemory and MmFreeSystemMemory

ergo720:
Same as https://github.com/Cxbx-Reloaded/xbox_kernel_test_suite/pull/81


RadWolfie:
Difference is the named variable string has been stripped down to only the name itself.

Approved

PR #89, opened 2023-12-28
Add RtlTimeFieldsToTime and RtlTimeToTimeFields tests

ergo720:
I noticed this in Cxbx-reloaded, where it seems that the implementation of this function is wrong. I believe that the function is supposed to return the time converted in 100 ns intervals, while now it seems it's only returning the ms elapsed from the Windows epoch time (January 1st, 1601). So I added these tests to confirm my suspicion. It's a draft for now because I haven't tested this yet on actual hardware.
EDIT: confirmed as working on hardware

PR #90, opened 2023-12-28
Fix QuadPart value output cause hang

RadWolfie:
The issue was discovered from making further tests into #89 pull request.

Plus updated Buffer's length to 512 bytes long appropriately.

### DIFF
--- a/src/assertions/defines.h
+++ b/src/assertions/defines.h
@@ -13,8 +13,10 @@
 #define GEN_CHECK_EX(check_var, expected_var, varname, func_line) \
     if(check_var != expected_var) { \
         print( \
-            "  ERROR(line %d): Expected %s = 0x%x, Got = 0x%x", \
-             func_line, varname, expected_var, check_var \
+            ((sizeof(check_var) > 4) ? \
+            "  ERROR(line %d): Expected %s = 0x%llx, Got = 0x%llx" : \
+            "  ERROR(line %d): Expected %s = 0x%x, Got = 0x%x") \
+            , func_line, varname, expected_var, check_var \
         ); \
         test_passed = 0; \
     }
@@ -23,8 +25,10 @@
 #define GEN_CHECK_RANGE_EX(check_var, expected_var, size, varname, func_line) \
     if(check_var < expected_var || check_var > expected_var + size) { \
         print( \
-            "  ERROR(line %d): Expected range %s = 0x%x-0x%x, Got = 0x%x", \
-            func_line, varname, expected_var, expected_var + size, check_var \
+            ((sizeof(check_var) > 4) ? \
+            "  ERROR(line %d): Expected range %s = 0x%x-0x%llx, Got = 0x%llx" : \
+            "  ERROR(line %d): Expected range %s = 0x%x-0x%x, Got = 0x%x") \
+            , func_line, varname, expected_var, expected_var + size, check_var \
         ); \
         test_passed = 0; \
     }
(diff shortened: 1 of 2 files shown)


src/util/output.c
```diff
@@ -14,9 +14,9 @@ static HANDLE output_filehandle = INVALID_HANDLE_VALUE;
 void print(char* str, ...)
 {
     va_list args;
-    char buffer[500];
+    char buffer[512];
```
PatrickvL: Could a comment be added explaining why the buffer size needs to be 512? (Mentioning the circumstances when this size is relevant helps future maintenance if and when those circumstances change)

src/assertions/defines.h
```diff
@@ -12,31 +12,55 @@
 
 #define GEN_CHECK_EX(check_var, expected_var, varname, func_line) \
     if(check_var != expected_var) { \
-        print( \
-            "  ERROR(line %d): Expected %s = 0x%x, Got = 0x%x", \
-             func_line, varname, expected_var, check_var \
-        ); \
+        if (sizeof(check_var) > 4) { \
```
PatrickvL: Comparable code appears a few more times below; would it make sense to use a separate define for this, and use that in all applicable places? (The slight difference in the repeated code blocks below seems to be a label that could be passed into the separate define as an argument)

src/util/output.c
RadWolfie: length of 512 == 0x200
length of 500 == 0x1F4

Not sure why it's at 500 length, it's best to follow what we are really used to. ~~Although, below may could limit to 510 in order to append "\n" to the string. Which leave the last part as null terminated.~~ fixed

src/assertions/defines.h
RadWolfie: If you have suggestion on the how to do it. Feel free to tell me, otherwise I do not know what other alternative method in C standard language.

**EDIT:** I tried to look up better solutions to no avail. Although using ternary operator can reduce the our code size to use one call with same existing arguments.

src/util/output.c
PatrickvL: Okay, well if it was known exactly which code calls this function and with which inputs, a maximum required length could be derived from that. But if that's infeasible or otherwise deemed not worth the time, then just keep it like this.

However, if this gets changed, it might be nice to declare the length in a const, and use that for both the buffer allocation size and (reduced by 2) for the limit argument to the `vsnprintf` call

RadWolfie: > Okay, well if it was known exactly which code calls this function and with which inputs, a maximum required length could be derived from that. But if that's infeasible or otherwise deemed not worth the time, then just keep it like this.

Not possible, these are prefix format string with dynamic inputs. Yet widen the buffer length dynamically is too much effort to constant double checking for standard library's __printf call if we had allocated enough or not.

> However, if this gets changed, it might be nice to declare the length in a const, and use that for both the buffer allocation size and (reduced by 2) for the limit argument to the `vsnprintf` call

This is the only place we format the string from the buffer cache. I don't see the need for declare the length in a const? Technically we "could" lower it down to 256 length. Also, nxdk's debugPrint is internally limited to 512 length currently.

PatrickvL: > Not possible, these are prefix format string with dynamic inputs. Yet widen the buffer length dynamically is too much effort to constant double checking for standard library's __printf call if we had allocated enough or not.

This is only being called by this project, so it should be possible to determine all possible inputs.
But indeed, a dynamic buffer would be impractical.

> This is the only place we format the string from the buffer cache. I don't see the need for declare the length in a const? Technically we "could" lower it down to 256 length. Also, nxdk's debugPrint is internally limited to 512 length currently.

A const would ease maintenance, that's all.

I'm not going to press this line if thought further, as you seem set on keeping it like this.

And since I'm fine either way, I'll just approve it.

PatrickvL (approved):

src/util/output.c
RadWolfie: Are you referring to like enum type of variable to be used for initial variable and limited range to format the string?

If so, then very well. I will update it.

**EDIT:** Also, this number had not been changed for 6 years until now.

PR #94, opened 2024-01-30
Add Output Video Option

RadWolfie:
I got annoyed by no space in statements. So I made the fix format commit first.

Since open log file and read config file needs to be done first. Any print calls to screen are require to be disable first until read from config file. Then make the output video default to true if `disable-video` key was not given or set to respective value.

~~I'm currently mark this pull request as draft to verify team's approval to move forward with. Otherwise, I could make video output to always false which will require contributors/testers to use config file.~~

---
In `config.txt` file:
`disable-video=<boolean, 1 or 0 number only>`

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -1,7 +1,6 @@
-#include <hal/debug.h>
 #include <pbkit/pbkit.h>
-#include <hal/video.h>
-#include <hal/xbox.h>
+#include <hal/video.h> // for XVideoSetMode
+#include <hal/xbox.h> // for XReboot
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
@@ -30,26 +29,26 @@ int load_conf_file(char *file_path)
         FILE_ATTRIBUTE_NORMAL,
         NULL
     );
-    if(handle == INVALID_HANDLE_VALUE) {
+    if (handle == INVALID_HANDLE_VALUE) {
         print("Could not open config file '%s' for read", file_path);
         return -1;
     }
 
     DWORD file_size = GetFileSize(handle, NULL);
-    if(file_size == INVALID_FILE_SIZE) {
+    if (file_size == INVALID_FILE_SIZE) {
         print("ERROR: Could not get file size for %s", file_path);
         return -1;
     }
 
     char* buffer = (char *)malloc(file_size);
-    if(buffer == NULL) {
+    if (buffer == NULL) {
         print("Malloc failed for file_size %u", file_size);
         return -1;
     }
 
     DWORD bytes_read = 0;
     BOOL result = ReadFile(handle, buffer, file_size, &bytes_read, NULL);
-    if(result == 0 || bytes_read != file_size) {
+    if (result == 0 || bytes_read != file_size) {
         print("Read failed for config file. result = %d, read = %u", file_size, bytes_read);
         return -1;
     }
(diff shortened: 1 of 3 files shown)


src/main.c
```diff
@@ -18,6 +17,9 @@
 #define GIT_VERSION "unknown"
 #endif
 
+// from util/output.h file private use here only
```
ergo720:
```suggestion
// defined in util/output.h file, used privately here only
```

src/main.c
```diff
@@ -79,50 +85,56 @@ int load_conf_file(char *file_path)
...
-void main(void){
-
-    XVideoSetMode(640, 480, 32, REFRESH_DEFAULT);
+void main(void)
+{
+    vector_init(&tests_to_run);
+    if (!open_output_file("D:\\kernel_tests.log")) {
+        return;
+    }
+    // If load config fail, then do what need to be initialize.
```
ergo720:
```suggestion
    // If loading config file fails, then do what's needed to be initialize the video output below.
```

src/main.c
```diff
@@ -79,50 +85,56 @@ int load_conf_file(char *file_path)
...
+        // NOTE: Pre-init is set to FALSE because of video initialization doesn't occur until below.
+        output_video = TRUE;
+    }
 
-    switch(pb_init()){
-        case 0: break;
-        default:
+    if (output_video) {
+        XVideoSetMode(640, 480, 32, REFRESH_DEFAULT);
+        // If pb_init return non-zero, then something went wrong from nxdk's end.
```
ergo720:
```suggestion
        // If pb_init returns non-zero, then something went wrong from nxdk's end.
```

src/util/output.c
```diff
@@ -9,6 +9,8 @@
 #include "global.h"
 #include "util/output.h"
 
+BOOL output_video = FALSE; // NOTE: Must be default to FALSE until config.txt is loaded and after video initialization is done.
```
ergo720:
```suggestion
BOOL output_video = FALSE; // NOTE: Must be set to a default value of FALSE until config.txt is loaded and after video initialization is done.
```

ergo720:
Nit: the format commit should probably have been submitted in a separate PR, since it's not part of the video output option being added here

src/main.c
RadWolfie: fixed

src/main.c
RadWolfie: I removed this comment and converted into a function to centralize actual default values there. Reviewer(s) will need to re-review the new changes.

src/main.c
RadWolfie: fixed, also added "the" and "call" in the sentence.

src/util/output.c
RadWolfie: fixed, also corrected "after video initialization is done" to "before video initialization".

RadWolfie:
> Nit: the format commit should probably have been submitted in a separate PR, since it's not part of the video output option being added here

Yes, it could had been in a separate pull request. Although, it is only one file, main.c, had to be fixed for format in order to accept new changes from this pull request without conflict.

src/main.c
```diff
@@ -18,6 +17,15 @@
 #define GIT_VERSION "unknown"
 #endif
 
+// defined in util/output.h file, used privately here only
+extern BOOL output_video;
+// Initialize the actual default values here if the config file is either successfully loaded before reading inputs or failed to load.
+static void init_default_values()
+{
+    // NOTE: Pre-init is set to FALSE because of video initialization doesn't occur until after load the config file.
```
ergo720:
```suggestion
    // NOTE: Pre-init is set to FALSE because video initialization doesn't occur until after the config file is loaded.
```

src/main.c
```diff
@@ -18,6 +17,15 @@
 #define GIT_VERSION "unknown"
 #endif
 
+// defined in util/output.h file, used privately here only
+extern BOOL output_video;
+// Initialize the actual default values here if the config file is either successfully loaded before reading inputs or failed to load.
```
ergo720:
```suggestion
// Initialize the actual default values here if the config file is either successfully loaded before reading inputs or it failed to load.
```

RadWolfie: fixed

src/main.c
RadWolfie: fixed

PR #95, opened 2024-02-04
Additional Log Info Support

RadWolfie:
## What's already upstreamed:
- build hash

## From this pull request:
- [X] Submitter
- [X] bios/kernel version
- [X] Xbox hardware info flag (helpful to see detail of what kernel sets to)
- [X] PIC version (including console type prediction)
- Xbox hardware revision:
  - [X] GPU (PCI)
  - [X] MCPX (PCI)

## Undecided:
- [ ] Name and/or description (created pull request #98)

## What will not be included:
- Run timestamp
- Used emulator + version / build hash + relevant settings

---

For any missing details, such as hardware revision, please discuss in #6 ticket. Thanks!

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -26,6 +26,8 @@ static void init_default_values()
     output_video = TRUE;
 }
 
+static char *submitter = NULL;
+
 int load_conf_file(char *file_path)
 {
     print("Trying to open config file: %s", file_path);
@@ -83,6 +85,12 @@ int load_conf_file(char *file_path)
         if (strcmp("disable-video", current_key) == 0) {
             output_video = !strtoul(strtok(NULL, "\n"), NULL, 16);
         }
+        if (strcmp("submitter", current_key) == 0) {
+            char *value = strtok(NULL, "\n");
+            size_t length = strlen(value);
+            submitter = calloc(length + 1, sizeof(char));
+            strncpy(submitter, value, length);
+        }
     }
 
     free(buffer);
@@ -133,6 +141,11 @@ void main(void)
 
     print("Kernel Test Suite");
     print("build: " GIT_VERSION);
+    print("submitter: %s", (submitter ? submitter : ""));
+    if (submitter) {
+        free(submitter);
+        submitter = NULL;
+    }
     run_tests();
 
     vector_free(&tests_to_run);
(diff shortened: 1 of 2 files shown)


PatrickvL (approved):
LGTM

src/main.c
```diff
@@ -26,6 +27,8 @@ static void init_default_values()
     output_video = TRUE;
 }
 
+char *submitter = NULL;
```
ergo720: Nit: because `submitter` is only used in this file, it could be marked as `static`.

src/util/hardware.h
```diff
@@ -0,0 +1,74 @@
...
+        HalReadSMBusValue(0x20, 0x01, 0, &version);
+        pic_version[i] = (char)version;
+    }
+    pic_version[3] = '\0';
+}
+
+static const char* getConsoleType(char pic_version[4])
+{
+    // See https://xboxdevwiki.net/Xboxen_Info
+    if (pic_version[0] == 'D' || pic_version[0] == 'B') {
```
ergo720: The linked [website](https://xboxdevwiki.net/Xboxen_Info) also has information for many prototypes and pre-launch date boxes, which I think this code should also detect too. Currently, if `pic_version[0]` is 'D' or 'B' but it's not a devkit or a chihiro system, this code will return "UNKNOWN". There's also the issue that there is a known prototype where `pic_version[0]` is 'A', and the current code will not catch that. While I understand that the note for that entry says the result is from a "random poster", I still feel you could return a model string such as "EVT / DVT1 (uncertain)"

RadWolfie: I'm assuming `A__` prefix is alpha protocol type machine which is only used in-house experiment or early access to certain partners for maybe feedback and such. Though, we (including nxdk repo), don't support prototypes anyway. Yet... we could simply add `A__` prefix and have it return as "PROTOTYPE` string in case of build xbe for prototype will be supported in the future. (In summary, I'll add prototype check for 'A' prefix soon.)

Return model string isn't intended planned to do as I don't know where to detect it as for specific EVT/DVT#. Plus model string isn't console type which will not return it. And will require a new function to return model string according to hardware specs.

If you looked at X(MODE)2, they are actually debug / dev kits (including chihiro) hardware. Which is what `B__` and `D__` prefixes are. If the kernel flags doesn't set devkit nor chihiro, then technically something went wrong with the kernel or homebrew kernel/patcher. Which is why I let it pass through to return "UNKNOWN".

Issue #66, opened 2020-02-03
Consider switching to std::vector


ergo720:
Currently, the test suite is using a custom vector implementation provided in `vector.h` and `vector.cpp`. However, it's limited: for example, it doesn't have an erase function and it's hard-coded to use integer types for its elements. I believe that nxdk now offers a native `std::vector` implementation in nxdk-libcxx, as also evidenced by NevolutionX (an app built with nxdk) which is already using them here https://github.com/dracc/NevolutionX/blob/37b3a280cb5269168d1120dbfc370e0968ab3cb7/main.cpp#L33. To solve the above limitations, I suggest to switch to the `std::vector` implementation provided by nxdk.

RadWolfie:
It depends on how bad we need standard vector. As far I had see, only `tests_to_run` variable is using vector.

PR #96, opened 2024-02-10
Add exclude tests option

RadWolfie:
I recently realize exclude tests will be very useful for emulators to skip them if they are not compatible or aren't working by design. Exclude tests become necessary to verify other test results instead of manually inputting many individual tests, just to remove one or some tests.

---
In `config.txt` file:
`tests-exclude=<export hex number>[,...]`

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -34,6 +34,8 @@ static void init_default_values()
 
 static char *submitter = NULL;
 
+static vector tests_to_run;
+
 int load_conf_file(char *file_path)
 {
     print("Trying to open config file: %s", file_path);
@@ -110,14 +112,14 @@ static void run_tests()
         print("No Specific tests specified. Running all tests (Single Pass).");
         print("-------------------------------------------------------------");
         int table_size = ARRAY_SIZE(kernel_thunk_table);
-        for (int k=0; k<table_size; k++) {
+        for (int k = 0; k < table_size; k++) {
             kernel_thunk_table[k]();
         }
     }
     else {
         print("Config File Was Loaded. Only running requested tests.");
         print("-----------------------------------------------------");
-        for (int k=0; k<tests_to_run.size; k++) {
+        for (int k = 0; k < tests_to_run.size; k++) {
             kernel_thunk_table[vector_get(&tests_to_run, k)]();
         }
     }
--- a/src/main.c
+++ b/src/main.c
@@ -35,6 +35,7 @@ static void init_default_values()
 static char *submitter = NULL;
 
 static vector tests_to_run;
+static vector tests_exclude;
 
 int load_conf_file(char *file_path)
 {
@@ -90,6 +91,13 @@ int load_conf_file(char *file_path)
                 vector_append(&tests_to_run, strtol(current_test, NULL, 16));
             }
         }
+        if (strcmp("tests-exclude", current_key) == 0) {
+            char *current_test;
+            char *tests = strtok(NULL, NEWLINE_DELIMITER);
+            while ((current_test = strtok_r(tests, ",", &tests))) {
+                vector_append(&tests_exclude, strtol(current_test, NULL, 16));
+            }
+        }
         if (strcmp("disable-video", current_key) == 0) {
             output_video = !strtoul(strtok(NULL, NEWLINE_DELIMITER), NULL, 16);
         }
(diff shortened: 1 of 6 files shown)


RadWolfie:
Converted to draft pull request because #97 pull request is top priority. Once it is merged, this pull request can leave draft state.

src/main.c
```diff
@@ -103,30 +113,66 @@ int load_conf_file(char *file_path)
...
-        print("No Specific tests specified. Running all tests (Single Pass).");
+        print("No specific tests were requested. Running all tests (Single Pass).");
+        if (tests_exclude.size) {
+            int remainder_size = 0;
+            for (int i = 0; i < tests_exclude.size; i++) {
+                if (kernel_thunk_table_size > vector_get(&tests_exclude, i)) {
+                    remainder_size++;
+                }
+            }
+            print("%d test(s) will be exclude.", remainder_size);
```
PatrickvL: Typo : exclude > excluded

src/main.c
```diff
@@ -103,30 +113,66 @@ int load_conf_file(char *file_path)
...
+        }
         print("-------------------------------------------------------------");
-        int table_size = ARRAY_SIZE(kernel_thunk_table);
-        for (int k=0; k<table_size; k++) {
-            kernel_thunk_table[k]();
+        for (int k = 0; k < kernel_thunk_table_size; k++) {
+            run_test(k);
```
PatrickvL, line 146: Contrary to the code above, it appears as if no tests are actually excluded in this loop?

Nevermind, all good - the exclusion is in `run_test()`.

src/main.c
```diff
@@ -103,30 +113,66 @@ int load_conf_file(char *file_path)
...
+            for (int k = 0; k < tests_to_run.size; k++) {
+                int test_n = vector_get(&tests_to_run, k);
+                for (int i = 0; i < tests_exclude.size; i++) {
+                    if (test_n == vector_get(&tests_exclude, i)) {
+                        remainder_size++;
+                        break;
+                    }
+                }
+            }
+            print("%d test(s) will be exclude.", remainder_size);
```
PatrickvL: Same typo

src/main.c
```diff
@@ -103,30 +113,66 @@ int load_conf_file(char *file_path)
     return 0;
 }
 
+static void run_test(int test_n) {
+    for (int i = 0; i < tests_exclude.size; i++) {
```
PatrickvL, line 117: Since this can become an exponential-time check, it might be better to implement with a bit set instead of a vector, so that this search loop can be changed into a much faster index-based bit test.

src/main.c
```diff
@@ -103,30 +113,66 @@ int load_conf_file(char *file_path)
...
+}
+
 static void run_tests()
 {
     print("Random seed used is %u", seed);
     if (tests_to_run.size == 0) {
-        print("No Specific tests specified. Running all tests (Single Pass).");
+        print("No specific tests were requested. Running all tests (Single Pass).");
+        if (tests_exclude.size) {
+            int remainder_size = 0;
```
PatrickvL: Nit: this variable does not really contain a remainder size. It contains an exclusion count in this code block.

src/main.c
```diff
@@ -88,6 +91,13 @@ int load_conf_file(char *file_path)
                 vector_append(&tests_to_run, strtol(current_test, NULL, 16));
             }
         }
+        if (strcmp("tests-exclude", current_key) == 0) {
+            char *current_test;
+            char *tests = strtok(NULL, NEWLINE_DELIMITER);
+            while ((current_test = strtok_r(tests, ",", &tests))) {
+                vector_append(&tests_exclude, strtol(current_test, NULL, 16));
```
PatrickvL, line 98: Would it be possible to alternatively specify tests to exclude _by name_ instead of (or in addition to) by number?

src/main.c
```diff
@@ -103,30 +113,66 @@ int load_conf_file(char *file_path)
...
+    for (int i = 0; i < tests_exclude.size; i++) {
+        // If a match is found in the exclusion list, then we skip the test.
+        if (test_n == vector_get(&tests_exclude, i)) {
+            test_n = -1;
+            break;
+        }
+    }
+    // Skip the test if test_n is a negative number.
+    if (test_n >= 0) {
+        kernel_thunk_table[test_n]();
```
PatrickvL, line 126: `kernel_thunk_table` is not actually a kernel thunk table, right? It is an array of kernel API test functions, so perhaps it's clearer to update its name accordingly.

PatrickvL (changes requested):
Some remarks to ponder...

src/main.c
RadWolfie: Technically, you are correct. I think I did not rename the variable. And I may not know a better wording for it. Unless it's better to refer as... `kernel_api_tests` variable?

src/main.c
RadWolfie: That will be a feature request and will be able to support in a separate branch. Which I had reworked the global tests to remove user error mistakes. Plus no longer need to manually add hex value and name of api strings for each tests. So, like I said, your suggestion is a feature request.

Currently, it's following same way for tests input.

src/main.c
RadWolfie: Hmm my thought are exclude is present and excluded would be past tense. Except grammar check says excluded because of "be". I will fix it.

src/main.c
RadWolfie: It's best to keep the wording different since `tests_exclude.size` and `(tests_)exclude_size` are too identical to cause confusion. Though, `exclude_count` can work as a replacement.

src/main.c
RadWolfie: integer (4 bytes) can only hold up to 32 bits and long long (8 bytes) can hold up to 64 bits. We have 378 kernel apis which need 6 of long long values. Plus this file is in C language which doesn't add such flexibility to extend such support.

Meaning we would have to write a utility to able support your suggestion. Unless there's something I do not know about? I rather avoid using third-party library nor create our own library for this type of complex.

PatrickvL: Maybe https://en.cppreference.com/w/cpp/utility/bitset helps

EDIT: scratch that. You already wrote this is C, not C++.

But, a bitset could be created, laid on top of a vector. Or if wasting memory doesn't matter much, each vector index could be associated with a test (when the int data is used as a boolean value)

RadWolfie: Yes, if this file was using C++ language, then I could had used the standard library's bitset class. Except it is not the main objective for this pull request. ergo720 did made a ticket #66 for suggestion change custom library to std::vector. This main.c file could change to C++ language to take advantage of C++ standard library usage.

Plus we also need to be sure that nxdk's custom libcxx can support bitset.

Keep in mind that nxdk's custom libcxx/pbclib are behind from upstream.

src/main.c
PatrickvL: Is this rename going to be part of this pull request, or in a later one?

RadWolfie: Later one, you'll know which one eventually. 😉

RadWolfie:
Please mark conversations resolved if they are properly fixed. As for other questions/suggestion require later pull request(s), please keep them open. 🙂

PatrickvL (approved):

src/main.c
RadWolfie: resolved in #99 pull request.

src/main.c
RadWolfie: Transfer into ticket #100

src/main.c
RadWolfie: Transfer to ticket #101

PR #97, opened 2024-02-11
HOTFIX: Fix line reader to support file saved from all platforms

RadWolfie:
I had some sort of issue with the submitter field from #95 pull request. And when I started to work on adding the name option field. It wasn't able to create the custom log file. Which led me back to the submitter field issue I had. I didn't check the log file in Notepad++ which exposed `\r` character leaked in the submitter field output. I recently realized strtok calls actually need `"\r\n"` instead of `"\n"` to work with all platforms' files to read from.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -18,6 +18,11 @@
 #define GIT_VERSION "unknown"
 #endif
 
+// To read line by line properly from all platforms' line-endings with their text editor software.
+// Newline delimiter reader requires "\r\n" to find whichever character is found first.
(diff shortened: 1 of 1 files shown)


PatrickvL (approved):
Seems risk free. Perhaps it's useful to add a comment that explains why both characters are required, but if that's too much hassle, just ignore this.

LGTM otherwise

RadWolfie:
> Perhaps it's useful to add a comment that explains why both characters are required...

Added comment to explained the reason.

PatrickvL:
Splendid!

PatrickvL (approved):

PR #98, opened 2024-02-19
Add name option in a separate file

RadWolfie:
Instead of add name option in config.txt file, I decided to use it in name.txt file for the following reasons:
1) Ability not need to re-edit config.txt file for each machine every time. This way, config file can be edit once and send to all machines to run the tests to get results back for the reason below.
2) Ability to name kernel_tests.log to a custom suffix name log. i.e. without name.txt file would be `kernel_tests.log` by default. With name.txt file will use `kernel_tests-<name>.log`. This allow for testers to download from all machines into same folder than download individual log files renamed to what machine was used.
3) Profit??

---
In `name.txt` file:
`<name (first line only)>`

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -33,6 +33,56 @@ static void init_default_values()
 }
 
 static char *submitter = NULL;
+#define LOG_NAME_PATH_DEFAULT "D:\\kernel_tests"
+#define LOG_FILE_EXT ".log"
+static const char* name_log_format = LOG_NAME_PATH_DEFAULT "-%s" LOG_FILE_EXT;
+static char* name_log = NULL;
+
+void load_name_file(const char* file_path)
+{
+    HANDLE handle = CreateFile(
+        file_path,
+        GENERIC_READ,
+        FILE_SHARE_READ,
+        0,
+        OPEN_EXISTING,
+        FILE_ATTRIBUTE_NORMAL,
+        NULL
+    );
+    if (handle == INVALID_HANDLE_VALUE) {
+        return;
+    }
+
+    DWORD file_size = GetFileSize(handle, NULL);
+    if (file_size == INVALID_FILE_SIZE) {
+        return;
+    }
+
+    char* buffer = (char *)malloc(file_size);
+    if (buffer == NULL) {
+        return;
+    }
+
+    DWORD bytes_read = 0;
+    BOOL result = ReadFile(handle, buffer, file_size, &bytes_read, NULL);
+    if (result == 0 || bytes_read != file_size) {
+        return;
+    }
+
+    CloseHandle(handle);
+
+    char *line;
+    char *rest = buffer;
+    if ((line = strtok_r(rest, NEWLINE_DELIMITER, &rest))) {
+        size_t length = strlen(line);
+        if (length) {
+            name_log = calloc(length + 1, sizeof(char));
+            strncpy(name_log, line, length);
+        }
+    }
+
+    free(buffer);
+}
 
 static vector tests_to_run;
 static vector tests_exclude;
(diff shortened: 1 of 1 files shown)


src/main.c
```diff
@@ -33,6 +33,57 @@ static void init_default_values()
...
+    if ((line = strtok_r(rest, NEWLINE_DELIMITER, &rest))) {
+        size_t length = strlen(line);
+        if (length) {
+            name_log = calloc(length + 1, sizeof(char));
+            strncpy(name_log, line, length);
+        }
+    }
+
+    free(buffer);
+    return;
```
ergo720: Nit: unnecessary `return` statement at the end of this function.

RadWolfie: Oh, the function was borrowed from `load_conf_file` function without value returns and prints. I will fix it soon.

src/main.c
```diff
@@ -33,6 +33,56 @@ static void init_default_values()
...
+
+    char* buffer = (char *)malloc(file_size);
+    if (buffer == NULL) {
+        return;
+    }
+
+    DWORD bytes_read = 0;
+    BOOL result = ReadFile(handle, buffer, file_size, &bytes_read, NULL);
+    if (result == 0 || bytes_read != file_size) {
+        return;
```
PatrickvL, line 70: Buffer is leaking here

src/main.c
```diff
@@ -33,6 +33,56 @@ static void init_default_values()
...
+        0,
+        OPEN_EXISTING,
+        FILE_ATTRIBUTE_NORMAL,
+        NULL
+    );
+    if (handle == INVALID_HANDLE_VALUE) {
+        return;
+    }
+
+    DWORD file_size = GetFileSize(handle, NULL);
```
PatrickvL, line 56: Since only the first line will be read from this file, it could be wise to safeguard against too large file sizes here (imagine the input is a multi-GiB file...)

src/main.c
```diff
@@ -33,6 +33,56 @@ static void init_default_values()
...
+    }
+
+    CloseHandle(handle);
+
+    char *line;
+    char *rest = buffer;
+    if ((line = strtok_r(rest, NEWLINE_DELIMITER, &rest))) {
+        size_t length = strlen(line);
+        if (length) {
+            name_log = calloc(length + 1, sizeof(char));
```
PatrickvL, line 80: Best not access a global variable here. Instead make `name_log` a return value

src/main.c
```diff
@@ -173,7 +223,16 @@ void main(void)
...
+    char* output_file_name = "D:\\kernel_tests.log";
+    // If name_log buffer is allocated, then we know it does have actual input.
+    if (name_log) {
+        size_t name_log_length = strlen(name_log);
+        name_log_length += strlen(name_log_format) - 2; // exclude %s
+        output_file_name = calloc(name_log_length + 1, sizeof(char));
+        snprintf(output_file_name, name_log_length, name_log_format, name_log);
+    }
+    if (!open_output_file(output_file_name)) {
         return;
```
PatrickvL, line 238: `output_file_name` is leaking here

src/main.c
```diff
@@ -173,7 +223,16 @@ void main(void)
 {
     vector_init(&tests_to_run, kernel_thunk_table_size);
     vector_init(&tests_exclude, 20);
-    if (!open_output_file("D:\\kernel_tests.log")) {
+    load_name_file("D:\\name.txt");
```
PatrickvL, line 228: When `load_name_file()` is changed into returning the `name_log` value, that variable no longer has to be global-scoped and thus can become a local variable (which can be declared in this same line where it would be assigned the return value from `load_name_file()`).

PatrickvL, line 228: Instead of going through all this trouble of reading a "name" file just to retrieve a `name_log`, why isn't this just read from the command line arguments? (Or is that feature not available on Xbox - I don't remember that detail...)

PatrickvL (commented):
Some remarks

src/main.c
RadWolfie: Xbox is not a desktop, none of the consoles to this day are desktops. So, there is no such thing as command line interface. Internally? Yes, there are arguments support designed from one app to another app. The difference is they are not exposed to users. Unless, of course, the main app is a development tool with cli arguments exposed.

In summary, it will require create such app to pass an argument and read from `LaunchDataPage` export. Plus it's best not to rely on third-party app to supply such info.

src/main.c
RadWolfie: And so should for config.txt reader. Except we don't expect this type of thing to happen from contributors/testers. This tool isn't meant for general users.

EDIT: In summary, won't fix.

src/main.c
RadWolfie: The issue also exists in config file reader too.

src/main.c
RadWolfie: Technically, `name_log` is bind to only main.c file hence the `static` assigned. From my view, it is a local (global) variable just like submitter, tests_to_run, and tests_exclude variables do.

Who knows, maybe there will be additional feature if there's a request for it. 🤷‍♂️

EDIT: In summary, won't fix.

src/main.c
RadWolfie: App is terminated here, not an issue and does not cause memory leak into reboot dashboard. Only contiguous memory allocation can cause memory leak.

EDIT: In summary, won't fix.

PR #102, opened 2024-02-25
Update README file

RadWolfie:
Major updates to the readme file:
- Fixed headers to reduce font size a bit.
- Added license and CI build state badges at the top.
- Added all available options can be done in config and name files with details.
- Updated all of the outdated links.
- Updated top description to no longer refer to Cxbx-Reloaded only. This xbox kernel test suite tool can be tested for any homebrew kernels. Plus a link to wiki page for homebrew kernel creators to add themselves without relying on pull requests and not to bloat the readme file for current and future homebrew kernels.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,41 +1,61 @@
 # xbox kernel test suite
-Xbox kernel APIs tester written using the opensoure nxdk
+[![License: GPL v2](https://img.shields.io/badge/License-GPL%20v3-blue.svg)](https://github.com/Cxbx-Reloaded/xbox_kernel_test_suite/blob/master/LICENSE)
+[![GitHub CI](https://github.com/Cxbx-Reloaded/xbox_kernel_test_suite/actions/workflows/ci.yml/badge.svg?branch=master)](https://github.com/Cxbx-Reloaded/xbox_kernel_test_suite/actions/workflows/ci.yml?query=branch%3Amaster)
 
-This is a tool for testing xbox kernel apis, in particular CxBx-Reloaded kernel implementation.
+Xbox kernel APIs tester written using the open-source nxdk.
 
-# HOW TO BUILD:
-All you need is nxdk. You can get it here: https://github.com/xqemu/nxdk
+This is a tool for testing the xbox kernel APIs on the hardware to verify tests are passing. Plus we have a dedicated wiki page for [homebrew kernels](https://github.com/Cxbx-Reloaded/xbox_kernel_test_suite/wiki/List-of-Homebrew-Kernels-Tested) have used this tool to test their kernel implementation.
 
-Here is a setup guide: https://github.com/xqemu/nxdk/wiki/Getting-Started
+## HOW TO BUILD:
+All you need is nxdk. You can get it here: https://github.com/XboxDev/nxdk
 
-# CONFIGURATION FILE:
-The configuration file should be called _**"config.txt"**_ and should be placed in the same directory of the xbe.
+Here is a setup guide: https://github.com/XboxDev/nxdk/wiki/Getting-Started
 
-This is an example of config.txt:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,41 +1,61 @@
...
-Xbox kernel APIs tester written using the opensoure nxdk
+[![License: GPL v2](https://img.shields.io/badge/License-GPL%20v3-blue.svg)](https://github.com/Cxbx-Reloaded/xbox_kernel_test_suite/blob/master/LICENSE)
+[![GitHub CI](https://github.com/Cxbx-Reloaded/xbox_kernel_test_suite/actions/workflows/ci.yml/badge.svg?branch=master)](https://github.com/Cxbx-Reloaded/xbox_kernel_test_suite/actions/workflows/ci.yml?query=branch%3Amaster)
 
-This is a tool for testing xbox kernel apis, in particular CxBx-Reloaded kernel implementation.
+Xbox kernel APIs tester written using the open-source nxdk.
 
-# HOW TO BUILD:
-All you need is nxdk. You can get it here: https://github.com/xqemu/nxdk
+This is a tool for testing the xbox kernel APIs on the hardware to verify tests are passing. Plus we have a dedicated wiki page for [homebrew kernels](https://github.com/Cxbx-Reloaded/xbox_kernel_test_suite/wiki/List-of-Homebrew-Kernels-Tested) have used this tool to test their kernel implementation.
```
ergo720:
```suggestion
This is a tool for testing the xbox kernel APIs on the hardware to verify their workings. In addition, we have a dedicated wiki page for [homebrew kernels](https://github.com/Cxbx-Reloaded/xbox_kernel_test_suite/wiki/List-of-Homebrew-Kernels-Tested) that have used this tool to test their kernel implementation.
```

README.md
```diff
@@ -1,41 +1,61 @@
...
+Here is a setup guide: https://github.com/XboxDev/nxdk/wiki/Getting-Started
 
-This is an example of config.txt:
+## CONFIGURATION FILE:
+> [!NOTE]
+> The configuration file should be called _**"config.txt"**_ and should be placed in the same directory of the xbe.
 
-```
-seed=5
+The following list of options can be used inside config.txt file:
```
ergo720:
```suggestion
The following list of options can be used inside the config.txt file:
```

README.md
```diff
@@ -1,41 +1,61 @@
...
+- `seed` = `<hexadecimal (support up to unsigned int max)>`
```
ergo720: I'd suggest to be explicit here and state that the maximum allowed value for the seed is 0xFFFFFFFF, just like it's done below.

README.md
```diff
@@ -1,41 +1,61 @@
...
+- `tests` = `<hexadecimal (support up to 17A)>[,...]`
+- `tests-exclude` = `<hexadecimal (support up to 17A)>[,...]`
+- `disable-video` = `<boolean>`[^1]
 
-tests=1,25,3,F
-```
+[^1]: boolean value can be 1 or 0
```
ergo720, line 24: This is not true. The code for the disable-video option is using a `if` statement to check if it should output to the video. This means that any value other than zero will not output, not just one like this line claims.

RadWolfie: Technically, 1-9A-F characters will be read as "true" which then will make video disable due to inverter. If no input or zero will be treat as video enable by default.

But this is more or less a guideline for readers should follow properly.

README.md
RadWolfie: fixed

README.md
RadWolfie: fixed

README.md
RadWolfie: fixed ~~, but replaced "their workings" to "they're working".~~

PR #104, opened 2024-02-26
Implement object type tests

RadWolfie:
The following tests has been implement:
- (010) ExEventObjectType
- (016) ExMutantObjectType
- (01E) ExSemaphoreObjectType
- (01F) ExTimerObjectType
- (040) IoCompletionObjectType
- (046) IoDeviceObjectType
- (047) IoFileObjectType
- (0F0) ObDirectoryObjectType
- (0F9) ObSymbolicLinkObjectType
- (103) PsThreadObjectType

All of the tests above have performed checks on member variables. Plus use relative calls purpose for their respective object creation to verify their object creations do match one of the above object types.

The tests had been passed on the hardware.

> [!NOTE]
> Cxbx-Reloaded will crash for:[^1]
> - ExEventObjectType
> - ExMutantObjectType
> - ExSemaphoreObjectType
> - ExTimerObjectType (doesn't crash but it may will)
> - IoCompletionObjectType
> - IoDeviceObjectType (doesn't crash but it may will)
> - IoFileObjectType
> - ObDirectoryObjectType (doesn't crash but it may will)
> - ObSymbolicLinkObjectType (doesn't crash but it may will)
> 
> PsThreadObjectType will not cause a crash due to xbox object implementation is completed and passed the test.

[^1]: The reason for these crashes is due to not being implemented or currently using the host's objects which are exposed outside of the xbox kernel environment is forbidden.

### DIFF
--- a/src/assertions/object_type.h
+++ b/src/assertions/object_type.h
@@ -1,0 +1,40 @@
+#include <xboxkrnl/xboxkrnl.h>
+
+#include "util/output.h"
+#include "defines.h"
+
+static BOOL assert_object_type_ex(POBJECT_TYPE object_type,
+                                  ULONG pool_tag,
+                                  BOOL has_close,
+                                  BOOL has_delete,
+                                  BOOL has_parse,
+                                  int func_line)
+{
+    ASSERT_HEADER;
+
+    BOOLEAN is_address_valid;
+    GEN_CHECK_EX(object_type->AllocateProcedure, ExAllocatePoolWithTag, ".AllocateProcedure", func_line);
+    GEN_CHECK_EX(object_type->FreeProcedure, ExFreePool, ".FreeProcedure", func_line);
+    // NOTE: Some object type has Close/Delete/Parse functions assigned.
+    GEN_CHECK_EX(object_type->CloseProcedure != NULL, has_close, ".CloseProcedure", func_line);
+    if (has_close) {
+        is_address_valid = MmIsAddressValid(object_type->CloseProcedure);
+        GEN_CHECK_EX(is_address_valid, TRUE, ".CloseProcedure valid", func_line);
+    }
+    GEN_CHECK_EX(object_type->DeleteProcedure != NULL, has_delete, ".DeleteProcedure", func_line);
+    if (has_delete) {
+        is_address_valid = MmIsAddressValid(object_type->DeleteProcedure);
+        GEN_CHECK_EX(is_address_valid, TRUE, ".DeleteProcedure valid", func_line);
+    }
+    GEN_CHECK_EX(object_type->ParseProcedure != NULL, has_parse, ".ParseProcedure", func_line);
+    if (has_parse) {
+        is_address_valid = MmIsAddressValid(object_type->ParseProcedure);
+        GEN_CHECK_EX(is_address_valid, TRUE, ".ParseProcedure valid", func_line);
+    }
+    // TODO: DefaultObject field require check against the object's address for some object types.
+    GEN_CHECK_EX(object_type->PoolTag, pool_tag, ".PoolTag", func_line);
+
+    return test_passed;
+}
+#define assert_object_type(object_type, pool_tag, has_close, has_delete, has_parse) \
+    assert_object_type_ex(object_type, pool_tag, has_close, has_delete, has_parse, __LINE__)
--- a/src/assertions/object_type.h
+++ b/src/assertions/object_type.h
@@ -50,8 +50,9 @@ static BOOL assert_object_header_type_ex(POBJECT_TYPE object_type,
     GEN_CHECK_EX(type_object != NULL, TRUE, "type_object", func_line);
     if (NT_SUCCESS(status)) {
         POBJECT_HEADER event_object_header = OBJECT_TO_OBJECT_HEADER(type_object);
-        // NOTE: DefaultObject's relative offset or address (above 0x80000000) are private.
-        //       Since they are private, we can only verify if the address is valid by call
+        // NOTE: DefaultObject's relative offset or address (either within kernel executable
+        //       or allocated by ObCreateObject) is not exposed through kernel APIs.
+        //       We can only verify if the absolute address is valid with a call to the
         //       MmIsAddressValid function.
         BOOLEAN is_address_valid;
         if ((LONG_PTR)object_type->DefaultObject >= 0) {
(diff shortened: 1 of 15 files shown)


src/assertions/object_type.h
```diff
@@ -0,0 +1,72 @@
...
+
+#include "util/output.h"
+#include "defines.h"
+
+static BOOL assert_object_type_ex(POBJECT_TYPE object_type,
+                                  ULONG pool_tag,
+                                  BOOL has_close,
+                                  BOOL has_delete,
+                                  BOOL has_parse,
+                                  int func_line)
```
ergo720, line 11: Instead of passing all the arguments to check to the function like this, you could create an array in `src/tests/io/suite/object_type.c` that holds all of them, and then this function would only receive an index/enumerator to the object to check.
```
struct object_info {
	POBJECT_TYPE object_type;
	ULONG pool_tag;
	BOOL has_close;
	BOOL has_delete;
	BOOL has_parse;
};

enum object_t {
	event_object_type = 0,
	// all the others
	object_max
};

// This could also use designated initializers instead of comments
object_info objects_to_test_info[object_max] = {
	{
		/* object_type */ &ExEventObjectType,
		/* pool_tag */ 'vevE',
		/* has_close */ FALSE,
		/* has_delete */ FALSE,
		/* has_parse */ FALSE
	}
};
```

src/assertions/object_type.h
```diff
@@ -0,0 +1,72 @@
...
+{
+    ASSERT_HEADER;
+
+    BOOLEAN is_address_valid;
+    GEN_CHECK_EX(object_type->AllocateProcedure, ExAllocatePoolWithTag, ".AllocateProcedure", func_line);
+    GEN_CHECK_EX(object_type->FreeProcedure, ExFreePool, ".FreeProcedure", func_line);
+    // NOTE: Some object type has Close/Delete/Parse functions assigned.
+    GEN_CHECK_EX(object_type->CloseProcedure != NULL, has_close, ".CloseProcedure", func_line);
+    if (has_close) {
+        is_address_valid = MmIsAddressValid(object_type->CloseProcedure);
```
ergo720, line 21: This call to `MmIsAddressValid` here is useless in my opinion. If the object has a procedure, then its function pointer points to an internal kernel function. Since that will reside in the kernel executable, and all its pages are mapped in virtual memory during initialization, then `MmIsAddressValid` will always return TRUE when the pointer is not nullptr.
Same for the two calls right below this code.

src/assertions/object_type.h
```diff
@@ -0,0 +1,72 @@
...
+                                         int func_line)
+{
+    ASSERT_HEADER;
+
+    PVOID type_object;
+    NTSTATUS status = ObReferenceObjectByHandle(object_handle, object_type, &type_object);
+    GEN_CHECK_EX(type_object != NULL, TRUE, "type_object", func_line);
+    if (NT_SUCCESS(status)) {
+        POBJECT_HEADER event_object_header = OBJECT_TO_OBJECT_HEADER(type_object);
+        // NOTE: DefaultObject's relative offset or address (above 0x80000000) are private.
```
ergo720: What do you mean here with "being private"? The XBE on the xbox runs in ring 0, so it should be able to access any memory it wants, as long as it doesn't cause a page fault, thus included those at addresses above 0x80000000.

src/tests/ex/suite/object_type.c
```diff
@@ -0,0 +1,77 @@
...
+{
+    TEST_BEGIN();
+
+    test_passed &= assert_object_type(&ExTimerObjectType, 'emiT', FALSE, TRUE, FALSE);
+    // NOTE: There's no such thing as CreateTimer for Windows, it is done through SetTimer.
+    ANSI_STRING obj_name;
+    RtlInitAnsiString(&obj_name, "ExTimerObjectType");
+    OBJECT_ATTRIBUTES obj_attr;
+    InitializeObjectAttributes(&obj_attr, &obj_name, OBJ_OPENIF, ObWin32NamedObjectsDirectory(), NULL);
+    // --------
```
ergo720, line 66: Is this line comment intentional or accidental?

src/assertions/object_type.h
RadWolfie: It is not useless. In fact, when tested against cxbxr failed this check as well. The intention is the memory does in fact is allocated and used by the kernel or one of the patched.

I had used `MmIsAddressValid` function for dump my xbox's kernel from allocated memory. When `MmIsAddressValid` call return false, it means there's nothing allocated in there and can be easily avoid attempt read into those memory space. Beside devices' unified memory range which will always true and unreliable there.

The purpose is ensure kernel's memory manager do return true as "something" did allocated there or whichever an exploit or homebrew kernel did something incorrectly. MmIsAddressValid cannot always return true for anything which had been confirmed on the hardware.

src/assertions/object_type.h
RadWolfie: Correct, except the kernel is compiled and stored inside xbox which is therefore "private". The only thing exposed are the kernel APIs. DefaultObject itself isn't in the kernel APIs.

If you had looked into the process of device creation where DefaultObject is only place of attachment to the device object or the like. Then you know there's no other places in kernel APIs exposed the DefaultObject to match against.

According to our discussion, it is VERY clear that it is not possible to retrieve its location directly other than analyzed process how the kernel get the address directly.

src/tests/ex/suite/object_type.c
RadWolfie: Intentional, see the note above.

src/assertions/object_type.h
RadWolfie: imo, having them done through parameters are better for education purpose. Or locally stored structure variable in the test's function, either way. Doing it internally local variable storage lose the education purpose for the test itself.


On the side note, C standard language always support `.var_name = assigned value` and C++20 standard and later. But C++20 isn't supported by nxdk's libcxx as it has partial support for C++17.

src/assertions/object_type.h
ergo720: So here "being private" means that the default object is not reachable from outside code because it's not exposed in any way. In that case I suggest to reword the comment though, because it's not clear to me as is. The first thing that I think when I see "being private" is that the memory is protected by the kernel and, as such, it cannot be accessed by the XBE.

src/assertions/object_type.h
RadWolfie: ergo720 said this can be resolved.

src/assertions/object_type.h
PatrickvL: An array would centralize the tested characteristics of all object types into one single place, which sentiment could be appreciated, but having to refer to them by index (while there's no other usage of such an index) introduces an extra indirection step that not only obfuscates the actual values used for the calls, but also requires careful maintenance (if anything would ever be changed, that is).

All in all, I'd go with the argument approach as it is right now in this pull request.

RadWolfie:
Rebased and added review remark.

src/assertions/object_type.h
RadWolfie: Redid the whole sentence for better clarification with some assistance from @ergo720.

PR #105, opened 2024-02-27
Replace custom vector to C++ std bitset class

### LINKED ISSUE #66: Consider switching to std::vector

### LINKED ISSUE #101: Optimize tests loop with std::bitset
Per @PatrickvL suggestion to improve faster check process for each tests that are excluded.

Although, I see std::bitset could be useful to fully replace custom vector since custom vector is only used in main.c file so far.

**EDIT:** However, we do need to check if nxdk's libcxx is able to support std::bitset.
**EDIT2:** Turns out std::bitset is supported.

RadWolfie:
Closes #101, #66

Per @PatrickvL suggestion to improve the loop process by using C++ standard library's bitset class which is supported by nxdk's C++ library. Doing so will require main.c rename to main.cpp and update makefile to use C++ compiler.

And updated the code to limit less than kernel_api_test_size can be set from input. Which then no longer needs a safety check inside for statement(s).

I have personally tested this with Cxbx-Reloaded to verify the exclusion(s) are working correctly and that all tests (except 0x77 test) are working as intended.


src/main.cpp
```diff
@@ -77,18 +79,18 @@ void load_name_file(const char* file_path)
...
         }
     }
 
     free(buffer);
 }
 
-static vector tests_to_run;
-static vector tests_exclude;
+static std::bitset<kernel_api_tests_size> tests_to_run;
+static std::bitset<kernel_api_tests_size> tests_exclude;
```
ergo720:
```suggestion
static std::bitset<kernel_api_tests_size> tests_to_exclude;
```

src/main.cpp
```diff
@@ -165,73 +173,46 @@ int load_conf_file(char *file_path)
...
-                for (int i = 0; i < tests_exclude.size; i++) {
-                    if (test_n == vector_get(&tests_exclude, i)) {
-                        exclude_count++;
-                        break;
-                    }
-                }
-            }
-            print("%d test(s) will be excluded.", exclude_count);
+        if (tests_exclude.any()) {
+            const auto found_tests_exclude = tests_to_run & tests_exclude;
```
ergo720:
```suggestion
            const auto tests_to_run_excluded = tests_to_run & tests_exclude;
```

src/main.cpp
```diff
@@ -140,14 +142,20 @@ int load_conf_file(char *file_path)
...
+                }
             }
         }
         if (strcmp("tests-exclude", current_key) == 0) {
             char *current_test;
             char *tests = strtok(NULL, NEWLINE_DELIMITER);
             while ((current_test = strtok_r(tests, ",", &tests))) {
-                vector_append(&tests_exclude, strtol(current_test, NULL, 16));
+                unsigned long value = strtoul(current_test, NULL, 16);
+                if (value < kernel_api_tests_size) {
```
PatrickvL, line 156: Would `tests_to_exclude[value]` be possible (and preferable) here?

PatrickvL (commented):
One small remark

src/main.cpp
RadWolfie: No, what you suggest is outside the limit of "size" which will cause access violation. Unless you're referring to one line below. Both `.set()` and `.reset()` are equivalent to using `var[value] = true/false;`, except I choose to use those bitset's functions directly.

If above are not what you're suggesting. Please clarify your question more clearly.

src/main.cpp
```diff
@@ -165,73 +173,46 @@ int load_conf_file(char *file_path)
...
+            tests_to_run &= ~tests_to_run_excluded;
+            print("%zu test(s) will be excluded.", tests_to_run_excluded.count());
         }
-        print("-----------------------------------------------------");
-        for (int k = 0; k < tests_to_run.size; k++) {
-            run_test(vector_get(&tests_to_run, k));
+    }
+    print("-------------------------------------------------------------");
+    for (int i = 0; i < kernel_api_tests_size; i++) {
+        if (tests_to_run.test(i)) {
```
PatrickvL, line 200: For unknown reason, my previous remark got misplaced;
What I intended was to ask if `if (tests_to_run.test(i))` could be slightly changed into `if (tests_to_run[i])`

src/main.cpp
RadWolfie: Continued conversation: https://github.com/Cxbx-Reloaded/xbox_kernel_test_suite/pull/105#discussion_r1526942957

src/main.cpp
RadWolfie: According to the bitset's doc:
Test caller can throw out of range exception. Whilst [] operator has undefined behavior for out of bounds. In my honest opinion, using the test function provided by bitset is better to use in case of the need to fix up either here (unlikely) or in the future codes using bitset to improve our skillset for using the standard bitset library.

bitset refs:
- [test(position/index)](https://en.cppreference.com/w/cpp/utility/bitset/test)
- [operator[]](https://en.cppreference.com/w/cpp/utility/bitset/operator_at)

PatrickvL: Fine by me.

PR #106, opened 2024-05-06
Add API name support for tests/tests-exclude options

### LINKED ISSUE #100: Add ability to use api name in tests/tests-exclude options
Per @PatrickvL request to have ability use api names in `tests`/`tests-exclude` options.

Having this option is only good for a few selection of api tests without need to know their hex index value. For longer api tests list, it is best to use hex index value.

At this time to add such ability require #99 pull request upstream. Afterward, this feature request will become possible to implement.

RadWolfie:
resolve #100

Like the title say, this pull request add feature to able use API names in the tests/tests-exclude options. The README file has been updated for input and example.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -90,6 +90,26 @@ void load_name_file(const char* file_path)
 static std::bitset<kernel_api_tests_size> tests_to_run;
 static std::bitset<kernel_api_tests_size> tests_to_exclude;
 
+constexpr size_t XBOXKRNL_API_NAME_STR_MIN_LENGTH = 6; // XcHMAC and RtlRip APIs are the shortest kernel names.
+unsigned long convert_test_api_input(char* test_str) {
+    // Trim space(s) before attempting to compare strings.
+    while(test_str[0] == ' ') {
+        test_str++;
+    }
+    if (strlen(test_str) >= XBOXKRNL_API_NAME_STR_MIN_LENGTH) {
+        for (unsigned long i = 0; i < kernel_api_tests_size; i++) {
+            // If the API name is a match, then return the index value.
+            if (_stricmp(test_str, kernel_api_tests[i].name) == 0) {
+                return i;
+            }
+        }
+        // If no match is found, then return a max value to skip add to the list.
+        return ULONG_MAX;
+    }
+    // Otherwise, we assumed the input is a hexadecimal string.
+    return strtoul(test_str, NULL, 16);
+}
+
 int load_conf_file(const char *file_path)
 {
     print("Trying to open config file: %s", file_path);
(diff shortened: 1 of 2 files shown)


PatrickvL (approved):
Looks good to me

src/main.cpp
```diff
@@ -90,6 +90,25 @@ void load_name_file(const char* file_path)
 static std::bitset<kernel_api_tests_size> tests_to_run;
 static std::bitset<kernel_api_tests_size> tests_to_exclude;
 
+unsigned long convert_test_api_input(char* test_str) {
+    // Trim space(s) before attempting to compare strings.
+    while(test_str[0] == ' ') {
+        test_str++;
+    }
+    if (strlen(test_str) > 4) {
```
ergo720: Instead of checking for an arbitrary string length, you can just attempt to call directly `strtoul` on the parsed string. If the test is specified with an API name, then the function will fail and will return either 0 or `ULONG_MAX`, and thus you can distinguish between the two cases.

RadWolfie: Hmm... when I made a sample test of:
```cpp
#include <climits>
#include <iostream>
#include <string>
 
int main()
{
    unsigned long value1 = strtoul("AB", NULL, 16);
    unsigned long value2 = strtoul("0xF", NULL, 16);
    unsigned long value3 = strtoul("10", NULL, 16);
    unsigned long value4 = strtoul("bsomething", NULL, 16);
    unsigned long value5 = strtoul("somethingb", NULL, 16);
    std::cout << "ULONG_MAX: " << ULONG_MAX << "\n";
    std::cout << "value1: " << value1 << "\n";
    std::cout << "value2: " << value2 << "\n";
    std::cout << "value3: " << value3 << "\n";
    std::cout << "value4: " << value4 << "\n";
    std::cout << "value5: " << value5 << "\n";
}
```
The output will return as:
```
ULONG_MAX: 18446744073709551615
value1: 171
value2: 15
value3: 16
value4: 11
value5: 0
```
which make `strtoul` function not entirely reliable. Any valid decimal/hexadecimal characters will be read and converted. Any invalid characters will return what's been found at beginning of the string.

The reason I have use `strlen` over 4 character lengths is normally known used for api name string. Below the length is known for hexadecimal (without `0x` prefix of course). If there's alternative method not to rely on `strtoul` return value, then I'm all for it.

As for `return kernel_api_tests_size;` which indeed should have use `ULONG_MAX` return.

RadWolfie: Also, smallest API name length so far is 6, `XcHMAC` and `RtlRip`.

ergo720: Ok, then maybe change the 4 to a macro name, to make it clearer what it means. Perhaps, something like `KRNL_STR_MIN_LENGTH`

PatrickvL (approved):

PR #110, opened 2025-09-02
Implement tests for (Nt|Ke)(Resume|Suspend)Thread APIs

RadWolfie:
The following kernel APIs are implemented for the tests:
- KeResumeThread
- KeSuspendThread
- NtResumeThread
- NtSuspendThread

The code has been optimized to be able to reuse the same purpose across Nt and Ke thread tests. However, `THREAD_DUO_EVENTS_` prefix macro could be used for other APIs if needed.

### DIFF
--- a/src/tests/ke/suite/thread.c
+++ b/src/tests/ke/suite/thread.c
@@ -1,0 +1,63 @@
+#include <xboxkrnl/xboxkrnl.h>
+
+#include "util/output.h"
+
+TEST_FUNC(KeAlertResumeThread)
+{
+    /* FIXME: This is a stub! implement this function! */
+}
+
+TEST_FUNC(KeAlertThread)
+{
+    /* FIXME: This is a stub! implement this function! */
+}
+
+TEST_FUNC(KeBoostPriorityThread)
+{
+    /* FIXME: This is a stub! implement this function! */
+}
+
+TEST_FUNC(KeDelayExecutionThread)
+{
+    /* FIXME: This is a stub! implement this function! */
+}
+
+TEST_FUNC(KeGetCurrentThread)
+{
+    /* FIXME: This is a stub! implement this function! */
+}
+
+TEST_FUNC(KeQueryBasePriorityThread)
+{
+    /* FIXME: This is a stub! implement this function! */
+}
+
+TEST_FUNC(KeResumeThread)
+{
+    /* FIXME: This is a stub! implement this function! */
+}
+
+TEST_FUNC(KeSetBasePriorityThread)
+{
+    /* FIXME: This is a stub! implement this function! */
+}
+
+TEST_FUNC(KeSetDisableBoostThread)
+{
+    /* FIXME: This is a stub! implement this function! */
+}
+
+TEST_FUNC(KeSetPriorityThread)
+{
+    /* FIXME: This is a stub! implement this function! */
+}
+
+TEST_FUNC(KeSuspendThread)
+{
+    /* FIXME: This is a stub! implement this function! */
+}
+
+TEST_FUNC(KeTestAlertThread)
+{
+    /* FIXME: This is a stub! implement this function! */
+}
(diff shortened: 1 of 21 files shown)


RadWolfie:
Here's the successful log that had been ran on xbox hardware: [kernel_tests-retail 1.6.log](https://github.com/user-attachments/files/22095852/kernel_tests-retail.1.6.log)

However, Cxbx-Reloaded has different failure results for the same repeated test. And sometimes may cause deadlock.

src/tests/ke/suite/thread.c
```diff
@@ -0,0 +1,250 @@
...
+    ULONG old_suspend_count, counter = 0;
+    NTSTATUS result;
+    DWORD result_wait;
+
+
+#if 0 // TODO: Could not catch invalid argument at this time.
+    if (suspend) {
+        BOOL catch = -1;
+        __try {
+            (void)KeSuspendThread(NULL);
```
ergo720, line 77: These tests are wrong. `KeSuspendThread` only raises an exception when the suspend count of the thread exceeds the maximum limit, and expects the thread object to always be valid (because it's supposed to have already been validated by `NtSuspendThread`). Passing an invalid object like this will cause a page fault in the kernel and there's no point in testing that.

src/tests/ke/suite/thread.c
```diff
@@ -0,0 +1,250 @@
...
+            catch = 0;
+        } __except(EXCEPTION_EXECUTE_HANDLER) {
+            catch = 1;
+        }
+        GEN_CHECK(catch, 1, "catch");
+    }
+    else {
+        BOOL catch = -1;
+        __try {
+            (void)KeResumeThread(NULL);
```
ergo720, line 95: `KeResumeThread` never raises any exceptions, so these tests are wrong too, and passing an invalid thread object to it is invalid too (same as `KeSuspendThread`).

src/tests/ke/suite/thread.c
RadWolfie: I assumed you missed few lines above?
`#if 0 // TODO: Could not catch invalid argument at this time.`

These codes aren't enabled by default due to unfinished work. The purpose of these tests are making sure they do break as seen in `GEN_CHECK(catch, 1, "catch");` line as Ke functions here are meant to check they are not safe-checking the argument inputs. Only the Nt functions from this pull request does have safe-checking the argument.

The whole purpose of this kernel test suite is to "brute force" validate that they work by design as intended by the official Xbox kernel.

> `KeSuspendThread` only raises an exception when the suspend count of the thread exceeds the maximum limit.

I did not know nor checked what was the real limit as I did not get to that point yet.

ergo720: > I assumed you missed few lines above?

I didn't, but it will become an issue when you eventually re-enable this code, so I thought to report it anyway.

> The purpose of these tests are making sure they do break as seen in GEN_CHECK(catch, 1, "catch"); line as Ke functions here are meant to check they are not safe-checking the argument inputs

I disagree. As a general rule of thumb, the test suite should only test conditions that the functions expect and can handle/recover from. Passing a nullptr here will cause a crash in the kernel, and anything that happens after that is undefined behavior. So, test results from such situations are invalid by definition in my opinion.

RadWolfie:
Updated disabled block to referred ticket task need to implement before able to re-update the code.

Added test for maximum suspend count noticed by ergo720.

New test reveal Cxbx-Reloaded's KeSuspendThread implement is wrong.
```
152 - KeSuspendThread: Tests Starting
  ERROR(line 160): Expected is_caught = 0x1, Got = 0xffffffff
  ERROR(line 161): Expected suspend_count_current = 0x7f, Got = 0xc000004b
  ERROR(line 162): Expected status_exception = 0xc000004a, Got = 0x0
  Test 'KeSuspendThread[#]' FAILED
 ```

